PR #6, opened 2021-09-15
conf(VETS-CPC-126): Multi-Step Dockerfile setup

SmoothWin:
changed the docker file in order to reduce the size of the image after docker composing (this removes the need to build the service before doing a docker compose up)


zellyk (approved):
LGTM

cgerard321:
**NOID is not allowed.** Everything must be done on a Jira Issue with its own named branch.  Do not merge this until you fix this.

PR #11, opened 2021-09-16
All fields that was to be added have been added

Louenes099:
Vets class has gained few more fields.
The database has been updated.
Data has been added.

### DIFF
--- a/vets-service/src/main/java/com/petclinic/vets/datalayer/Vet.java
+++ b/vets-service/src/main/java/com/petclinic/vets/datalayer/Vet.java
@@ -35,6 +35,54 @@ public class Vet {
     @NotEmpty
     private String lastName;
 
+    @Column(name = "email")
+    @NotEmpty
+    private String email;
+
+    @Column(name = "phone_number")
+    @NotEmpty
+    private String phoneNumber;
+
+    @Column(name = "resume")
+    @NotEmpty
+    private String resume;
+
+    @Column(name = "workday")
+    private String workday;
+
+    public String getEmail() {
+        return email;
+    }
+
+    public void setEmail(String email) {
+        this.email = email;
+    }
+
+    public String getPhoneNumber() {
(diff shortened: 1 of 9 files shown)


SmoothWin (approved):
Checked and all fields have been modified correctly, the respective fields only increased and with the increase, respective modifications to the insert only added more insert data.

SmoothWin (changes requested):
Make sure to add tests for the new model changes
Also I'd recommend to at least make one of the vet entity fields or type unique. For example I'd really recommend to put the email and the phone number entity as a unique value. This will facilitate when we make requests later on for when we will use the DTO versions of entities to do post requests because the only unique field we have right now is the ID and we normally don't pass row IDs to the client slide of an application.
 I'd also recommend adding another unique field called UUID. This will be used to do vet user target specified tasks such as for modifying vets or deleting vets.

amine-12 (dismissed):

SmoothWin:
Also I'd recommend to at least make one of the vet entity fields or type unique. For example I'd really recommend to put the email and the phone number entity as a unique value. This will facilitate when we make requests later on for when we will use the DTO versions of entities to do post requests because the only unique field we have right now is the ID and we normally don't pass row IDs to the client slide of an application.

SmoothWin:
> Make sure to add tests for the new model changes
Also I'd recommend to at least make one of the vet entity fields or type unique. For example I'd really recommend to put the email and the phone number entity as a unique value. This will facilitate when we make requests later on for when we will use the DTO versions of entities to do post requests because the only unique field we have right now is the ID and we normally don't pass row IDs to the client slide of an application.

SmoothWin (approved):

Louenes099:
image

SmoothWin:
Test for getting all vets is written ready to merge pull request

amine-12 (approved):

PR #13, opened 2021-09-18
feat(AUTH-CPC-95): Add user roles

Fube:
- Added roles persistence (repo and migration scripts)
- Hooked up roles to users
- Made UI for viewing, adding, and deleting roles
- Setup SpringSecurity config for `/roles` endpoint (needs ADMIN role, might switch this to an authority instead later)


JordanAlbayrak (approved):
It looks good (according to an untrained eye) ! I tested every test and they all passed. I also looked through your code to get an idea on how to get the UI to work and I think its finally making sense thanks for that, boy is AngularJS awesome.. am I right?

Trilikin21 (approved):
Tested the persistence and application tests , as well as verifying the UI implementation. Files changes make sense. Everything looks good from my part.

PR #16, opened 2021-09-22
feat/VETS-CPC-93_Add_Fields

Louenes099:


BunTymofiy (approved):

Louenes099:
image
image
image

SmoothWin (approved):
Reviewed changes and tested current branch on local device. The service works and the few a new column value for the vet entity was added in order to track if the vet is still in service (active).

PR #17, opened 2021-09-22
bug(AUTH-CPC-160): use role service instead of repo

Fube:
- Made sure RoleController calls RoleService and not RoleRepo directly
- Added JaCoCo for coverage reports and setup a 90% minimum coverage to pass tests
- Refactored each test type into its own file for readability and (my own) sanity
- Made sure to get 100% coverage


Trilikin21 (approved):
Looks good, not much else to say.

ArianneBonneau (commented):
good work

ArianneBonneau (approved):

PR #23, opened 2021-09-23
Feat(VETS-CPC-35):Display vet summary info

SmoothWin:
Finalized changes to the ui in order to display a modal on hover when the name of the vet is hovered.
The modal is also responsive.


SmoothWin:
Here is an example of the feature running
screenshot

Louenes099 (approved):
Too many empty files but seems as if they weren't modified otherwise seems okay

amine-12 (approved):

PR #24, opened 2021-09-23
feat(AUTH-CPC-105): clear text pwd for login

ArianneBonneau:
- Added the clear text pwd for login page


Fube (changes requested):
Why did you add a login service field?
There is, to my knowledge, no login specific service
It should be auth service, but even then you're overextended. That should be added by APIG

Fube (commented):
LGTM 🤙

Fube (approved):

JordanAlbayrak (approved):
Looks good to me! I didn't notice anything that would break main.

PR #26, opened 2021-09-23
bug(AUTH-CPC-102): Signup user

JordanAlbayrak:
Fixed issue with UserController using the UserRepo instead of the UserService, modified a related test aswell. Also changed "Agony" to "Roles" in the Roles UI page. Tested to make sure this doesn't break main.


Fube:
Fix your PR name to match the convention

Fube (changes requested):
Why does UserService take in a User?
Should it not be a DTO?

Trilikin21 (approved):
You fixed your controller, as well as your test.

Fube (approved):
LGTM 👌

Trilikin21 (approved):
Corrected the method parameters to accept a dto instead of a user. Good job

PR #29, opened 2021-09-24
(Feat/VETS CPC-33) Added the Display info page when clicked on the vet's name

Iana-Sha:
Added the Display info page when clicked on the vet's name


Louenes099 (approved):
Tried the code and it works

PR #31, opened 2021-09-24
Feat/vets cpc 33 display vet detailed info page

Iana-Sha:
Feature is done


Louenes099 (approved):
Everything works on my side

SmoothWin (approved):
All works Iana had the story running on docker and the application fully works with the new modal pop-up changes

PR #33, opened 2021-09-24
feat(BILL-CPC-74) : basic tests

Dratorz:
JUnit tests for CRUD

### DIFF
--- a/billing-service/build.gradle
+++ b/billing-service/build.gradle
@@ -40,4 +40,7 @@ dependencies {
 }
 test {
 	useJUnitPlatform()
+	testLogging {
+		events "passed", "skipped", "failed"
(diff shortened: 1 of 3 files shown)


Trim0500 (approved):
All's well in the code!

image_2021-09-25_195800

All the tests are present (with the exception of update since we agreed there's no need for it), the code is commented (Although I would suggest fleshing them out a bit more to let the reviewer know what key lines do) The naming convention for the PR is respected and the checks pass.

Lastly the removal of Lombok is smart as the dependency is likely to cause problems later with inconsistency.

partenersintime (approved):
Looks good. Tests run and github is not crying. This is epic.

PR #35, opened 2021-09-24
feat(AUTH-CPC-64) : email uniqueness

Trilikin21:
Change the createUser service to not accept emails already in use in the database.
Implemented email constraint.
Implemented test for the uniqueness of the email in the database and the constraints.
Removed unnecessary imports and throw exceptions in the test.


zellyk (approved):
Looks good. No comments in the code however :P

JordanAlbayrak (approved):
Looks good to me

PR #37, opened 2021-09-24
feat(VETS-CPC-90): Added backend data validation method to filter out bad inputs

Louenes099:
Testing has also been added to test bad data.
Everything works.


SmoothWin (changes requested):
All is good but test display names are too long. Display Names should be shorter and method names should not be should<something...something> but just be a short description of the method itself: for example method name and display for "shouldRegisterTheDataByFilteringOutTheTypos()" should instead be 
VetRegisterDataFilterAndValidationTest
and for the display name it should be no what it will do it should just be similar to the method name such as:
"Vet input submission cleanup and validation Test"
Change these method names and display names and I will accept the pull request

amine-12 (dismissed):

SmoothWin (changes requested):
And remove the system.out.println in line 70 and 102 inside the VetResourceTest.java file

SmoothWin (approved):
Changes are satisfactory the printlns are removed and display names modified/ added

amine-12 (approved):

PR #46, opened 2021-09-25
feat(APIG-CPC-136): update the dockerfile for api, created endpoints for update/delete/add http requests, and junit tests.

Kenny-star:
* Added routing for the CreateVisit() method
* Added routings for Add/Delete/Update visit in the API and created Jun… 
* Updated the API DockerFile


cjayneb (changes requested):
**Redact any of the changes that were made on files within the visits service.**

The following is regarding the **VisitServiceClient**:

 we need another method for getVisitsForPet which takes a single petId and returns a Flux of VisitDetails; the uri should be "/visits/{petId}". 

Then, the uri for createVisitForPet* should be "/owners/*/pets/{petId}/visits" for it to match our Postmapping in VisitResource. 

The upadeVisitForPet* should take a VisitDetails object and pass it in the uri which should be the same uri as create. 

The, deleteVisitForPet* should take a visitId and pass it to the uri which should be "/visits/{visitId}".

CreateVisit should return one VisitDetails object not Mono<Visits> because we are only creating on visit

^same thing for update visit

Im pretty sure deleteVisit will not return anything because our method for deleting a visit in VisitResource has void return type.

Kenny-star:
> **Redact any of the changes that were made on files within the visits service.**
> 
> The following is regarding the **VisitServiceClient**:
> 
> we need another method for getVisitsForPet which takes a single petId and returns a Flux of VisitDetails; the uri should be "/visits/{petId}".
> 
> Then, the uri for createVisitForPet* should be "/owners/*/pets/{petId}/visits" for it to match our Postmapping in VisitResource.
> 
> The upadeVisitForPet* should take a VisitDetails object and pass it in the uri which should be the same uri as create.
> 
> The, deleteVisitForPet* should take a visitId and pass it to the uri which should be "/visits/{visitId}".
> 
> CreateVisit should return one VisitDetails object not Mono because we are only creating on visit
> 
> ^same thing for update visit
> 
> Im pretty sure deleteVisit will not return anything because our method for deleting a visit in VisitResource has void return type.

Wouldn't it be the opposite for createVisitForPet and updateVisitForPet in turns of parameters to pass in? From my understanding, we essentially want to create visits using a VisitDetails object with a "petId-less" uri, then we'd use the update visits with a petId in the parameter and uri make the request for the specific data.

cjayneb:
> with

You are probably right for the create as long as the petId is in the VisitDetails object. But for update we can't only have a petId, we need a VisitDetails object or else our service won't know which fields to update.

cjayneb (commented):
@Kenny-star same comments as last time but we should talk on discord it will be simpler

PR #47, opened 2021-09-25
feat(VISE-CPC-66): Modified Add Visit to include a new Practitioner ID field

bthorng:
* Modified the Add Visit form to include a Select input which will also send a Practitioner's ID based on what is selected.
* Modified the SQL schemas and data to reflect the new Practitioner ID field.
* Added basic tests for the three layers.


cjayneb:
Will probably need to test more than the valid scenario but other than that looks good to me

cjayneb:
ALL tests will also have to be refactored to include the practitionerId once this gets pushed

lbains (approved):

NickEspitalier (dismissed):

codingbeast666 (approved):

lbains (approved):

PR #50, opened 2021-09-25
feat(VISE-CPC-116): Added ability to update a Visit 

lbains:
-Added the required PutMapping
-Updated the business layer to support updating a visit
-Added tests for the new update feature


cjayneb (commented):
I don't think you should have the option to set the visitId. Also the tempVisit in the VisitServiceImpl test for update is not necessary.

cjayneb (commented):
Again, should probably test more than one scenario but looks good atm

NickEspitalier (approved):

bthorng (dismissed):

bthorng (approved):

NickEspitalier (approved):

PR #56, opened 2021-09-26
Feat/(VISE CPC - 47) Implement delete capabilities for the visits

NickEspitalier:


cjayneb (commented):
Everything looks fine except you shouldn't be able to setId() in the Visit class, and the last repo test for deleting an empty visit does not assert the right thing. It should assert that all the visits which were saved in the setupdb() are still there. So assert that when you findByPetId, the size of the response from the repo is still 2.

cjayneb (dismissed):
**VisitsServiceImplTests**
whenVisitDoNotExist -> your vise object should have an id other than 1 or 2 if you actually want to test the deleteVisit method.

**PersistenceTests**
Is_Visit_Empty_Dont_Delete -> you have to assert that the repo still has two visits in it.

**VisitResourceTests**
whenValidVisitIdDeleteTheVisit -> should verify that visitService.delete(1) is called once

codingbeast666 (approved):
gg

bthorng (approved):

bthorng (approved):

codingbeast666 (approved):
lgtm

PR #62, opened 2021-09-26
Conf(BILL-73):configure billing service

LemireCoding:


Trim0500 (approved):
Everything looks good for merging!

image

All the tasks are accomplished and set up as per Jira documentation. I would advise for next time to separate the stories into 2 branches next time (although this time is kind of an exceptional situation due to improper backlog refinement from all of us). Additionally, put in comments for the tests to make sure that any reviewer from another team knows what's going on if the task calls for it.

jumpung (approved):
Everything looks good

Only thing that you might want to add is comments so that we can follow along afterwards

Dratorz (approved):
Everything is looking good! Naming convention are followed and I don't see any problem with the code itself. You might just want to add some comments so that others know what is going on.

PR #65, opened 2021-09-27
feat(vets-cpc-40): modify vet info

BunTymofiy:
Finished Modify feature


SmoothWin (approved):
Checked changes and tymofiy demoed the current state with the application and the curls work he also modified the dockerfile in order to support curl commands directly to the docker image

amine-12 (approved):

PR #66, opened 2021-09-27
feat(AUTH-CPC-13): chk password strength

MaxGrabs:
not fully tested but annotation is there.


ArianneBonneau (approved):
good start, but will need more work in sprint 2 (not 100% complete)

zellyk (approved):
lgtm!

PR #72, opened 2021-09-27
feat( VISE-CPC-99) : fixed visits html so it displays visits , added a visits controller

codingbeast666:
Previous visits would no longer display on the UI once we switched to 3 layer architecture .


cjayneb (approved):
LGTM

lbains (approved):
Nice

Funnyadd (approved):
lgtm

PR #78, opened 2021-09-27
feat(APIG-CPC-136): Create/Add visit

Kenny-star:
*Added routing for the CreateVisit() method
*Added routings for Add/Delete/Update visit in the API and created Jun…
*Updated the API DockerFile


cjayneb (commented):
Will obviously need to clean up unused methods later on and delete doesn't work. Also I can't approve and merge because we found out that main is broken (customer and billing docker containers keep exiting) so it would just make it harder to pinpoint the problem. Really wish I can approve it by tomorrow before noon.

cjayneb (approved):
LGTM

codingbeast666 (approved):
Blood, sweat and tears

PR #80, opened 2021-09-27
Feat(VETS-CPC-63): Enable Vet

Iana-Sha:
Finished enable vet, added the test and the ui.
UI will have to wait for APIG.
Curl commands work


Louenes099 (approved):
Everything works

BunTymofiy (approved):

PR #83, opened 2021-09-27
bug(AUTH-CPC-206): fix role controller from using a mapper

Trilikin21:
Made roleService to changes to the DTO and move the try catch method for catching an id not in the db inside the role service.

Test are also updated to reflect changes

[index.pdf](https://github.com/cgerard321/champlain_petclinic/files/7239705/index.pdf)


zellyk (approved):
Great work! Much needed.

Fube (approved):
LGTM 👍

JordanAlbayrak (approved):
Approved

PR #84, opened 2021-09-27
Feat(AUTH-CPC-214): add swagger

Trilikin21:
added swagger in the build.gradle for later use


JordanAlbayrak (approved):
Looks good to me! But I think you're asking a bit much by making us review all that code, maybe cut it down a little bit? Thanks

zellyk (approved):
LGTM👍

PR #85, opened 2021-09-28
Bug(AUTH-CPC-205): signup password hidden

JordanAlbayrak:
I simply fixed the bug where the signup password was not hidden. I also added some headers that might be removed or modified later on. No test necessary so  I will not post an imagine of JaCoCo.


zellyk (approved):
lgtm 👍

Trilikin21 (approved):
Looks good to me.

PR #93, opened 2021-09-29
bug(VISE-CPC-212): Fixed ID Setter and Removed Duplicate ID Setter

lbains:
- Put Mapping wasn't working properly due to there being an error in the Visit Class.
- Fixed the ID setter (the setter had int Id as an argument, meant to be int id)
- Removed the extra ID setter that was in the Visit Class


cjayneb (approved):
LGTM!

bthorng (approved):
LGTM 👍

PR #98, opened 2021-09-29
feat/CUST-CPC-287_Remove_Custodian

Anh-Toine:
Temporarily removing the Add Custodian feature as it is not working as expected.


ShiiroFX (dismissed):
Good work, hopefully we can fix this issue and bring it back later on. Code looks good, though you'll have to resolve the conflicts for the build to pass.

ShiiroFX (approved):
Nice work, this should make everything go a lot smoother now!

Louispoire (approved):
Perfection. I tested in out in IntelliJ and it work just fine! We can resume our testing!

PR #102, opened 2021-09-30
Feat(VETS-CPC-236): Vet search field

Louenes099:
Added search field for vet that allows user to search vet according to any field that the vet in question may have.


Louenes099:
image
image
image
image
image

Proof of the search field working and that it can search any field of the vets

amine-12 (approved):

BunTymofiy (approved):

PR #108, opened 2021-10-01
feat(BILL-CPC-249):Add customerId to Database and Bill Object

jumpung:
In this pull request I added customerBillId to everything that needed it

### DIFF
--- a/billing-service/src/main/java/com/petclinic/billing/datalayer/Bill.java
+++ b/billing-service/src/main/java/com/petclinic/billing/datalayer/Bill.java
@@ -14,6 +14,9 @@ public class Bill {
 
     private int billId;
 
+    @Column(name="customerId")
+    private int customerId;
+
     @Column(name = "visit_date")
     @Temporal(TemporalType.TIMESTAMP)
     @JsonFormat(pattern = "yyyy-MM-dd")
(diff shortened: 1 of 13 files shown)


Trim0500 (changes requested):
Can't accept this right now

Jira:

image

There is no user story put in, I don't know what this story is supposed to accomplish. There are no subtasks defined either. We can't know what the status of the story is.

The PR naming convention isn't followed, remember the format is feat(TEAMTAG-CPC-###): Desc

Lastly and credit to Justin, but the customerBillId should be a foreign key if we are to use Auth service's sessions

billing-service/src/main/java/com/petclinic/billing/datalayer/Bill.java
```diff
@@ -27,7 +30,9 @@
 
     public Bill(){}
 
-    public Bill(int billId, Date date, String visitType, double amount){
+    //int customerId ---> Add this to constructor maybe?
```
Trim0500, line 33: I think we can remove the comment out here and implement the field your considering. Do you agree?

LemireCoding (approved):
Comments were removed like asked. Good and ready to merge

Trim0500 (approved):
All changes have been addressed, I approve this message!

LemireCoding (approved):
Comments were removed. Everything looks good to go

PR #109, opened 2021-10-01
feat(VISE-CPC-194): Add exception handling

cjayneb:
- Added utils package to our service which contains http package and exceptions package
- http package contains the ControllerExceptionHandler and HttpErrorInfo class
- exceptions package contains the classes for InvalidInputException and NotFoundException
- Added tests for both exception classes (will add more)
- Excluded http package from jacoco report for debugging purposes


cjayneb:
PR is not ready for submission

PR #110, opened 2021-10-02
Feat(VETS-CPC-231): Front-End Data Validation

Louenes099:
Did the maximum of Data Validation possible within the front-end (vet-form.html).

Limited verification is doable with AngularJS.

Back-End verification covers more ground for the data clean/verification.

### DIFF
--- a/api-gateway/src/main/resources/static/scripts/vet-form/vet-form.controller.js
+++ b/api-gateway/src/main/resources/static/scripts/vet-form/vet-form.controller.js
@@ -13,7 +13,6 @@ angular.module('vetForm')
                 self.vet = resp.data;
             });
         }
-
         self.submitVetForm = function () {
             var id = self.vet.vetId;
             var req;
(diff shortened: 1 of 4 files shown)


Louenes099:
Everything works

Louenes099:
Front-End data validation was halfway added when the form was created

Louenes099:
image
image
image

Proof of everything working

SmoothWin (changes requested):
you have commited changes I see but we need examples of the implementation or logic sketch / diagram in order to show what your implementations did.

Would like you to add these to the pull request

Louenes099:
image

This image shows an input field for firstName with the requirements. The span is shown when the fields are improperly filled after the form submit button is clicked and the logic is being taken from the JS file. Right now it isn't working as the form is independent in its own way right now as it is missing components from the API-Gateway team.

Louenes099:
The previous comment still applies after the changes

Louenes099:
image
image
image
image

All necessary proof of the application running after changes

SmoothWin (dismissed):

SmoothWin (changes requested):
What is with you throwing exceptions inside the catch block
shouldn't they be thrown inside the try block.
change this please.

SmoothWin:
Never mind the first line comment u can keep the one in the catch block.
But in order for you to reach that catch block you will need to manually throw an exception of type HttpClientErrorException inside the try block in order to then to reach the catch block and reach your catch block or else it will never reach that block

Louenes099:
image
image

Louenes099:
image
image

Proof of everything working in terms of deploying and usage. Missing APIG's work to plug the submit button to the service so that all the input data be saved in the database.

Chloedunwoody (approved):
clean and functional, see no issues.

JordanAlbayrak (approved):
From what I can tell, everything seems fine. I approve it.

PR #115, opened 2021-10-03
feat(VISE-CPC-194): Add exception handling

cjayneb:
-     Added utils package to our service which contains http package and exceptions package
-     http package contains the ControllerExceptionHandler and HttpErrorInfo class
-     exceptions package contains the classes for InvalidInputException and NotFoundException
-     Added tests for both exception classes, HttpErrorInfo class, and ControllerExceptionHandler class (100% coverage)
-     Added simple headers for each class in our service
-     Added a check for empty visit description and duplicate visitId which throws and InvalidInputException in the business layer to show an example of exception handling

BUG REPORT

Although the InvalidInputException is handled gracefully during runtime, an IllegalStateException is still thrown.
I created a bug on Jira to investigate this problem and another problem that could be related: https://champlainsaintlambert.atlassian.net/browse/CPC-313?atlOrigin=eyJpIjoiNjRjMjYxYThmZjFhNGM5Zjg0MmNjNWEzOGM0M2VlYzgiLCJwIjoiaiJ9


bthorng (approved):
LGTM, we could also make other fields necessary when creating a visit such as practitionerId later on as well.

codingbeast666 (approved):
LGTM . Lots of tests are done ; we should check for redundancy at some point.

PR #120, opened 2021-10-03
feat(VISE-CPC-99):  Retrieve Vet Workdays from Select for Visits

codingbeast666:


cjayneb (approved):
You should add a clearer description, because this is definitely not backend. However, I approve of the changes.

Funnyadd (approved):
Lgtm

bthorng:
Looks good but, if you're selecting a practitioner wouldn't it show their names and not workdays? I need a little clarification on that.

Also for the select's value we should have it as the practitioner's ID, otherwise when the form sends a POST request, it's going to put the practitioner's workday instead of the practitionerId in the JSON.

For example right now:
`'practitionerId':'Monday, Thursday, Friday'`
As opposed to:
`'practitionerId':'215132'`

codingbeast666:
For now, I was asked by Adam to keep it simple and return only the workdays. This is mostly so that we can easily retrieve workdays and use them within the calendar. I commented a line next to it that will look more like the final format. That line contains practioner name , lastname and their workdays. 
@brandonthorng 
I changed the description. You are right. This is very very front-end 👍 
@cjayneb

PR #124, opened 2021-10-03
feat(VISE-CPC-265): Display more vet info when selected

bthorng:
* Added disabled input fields to the Add Visit that display the veterinarian's phone number, email, specialties and workdays
* The input fields are automatically filled with a vet's information when one is selected


codingbeast666 (approved):

Funnyadd (approved):
lgtm

PR #126, opened 2021-10-04
Feat(VISE-CPC-175): Vet availability input in Calendar UI

Funnyadd:
Added a feature to see the availabilities of the different vets depending on their workdays on the calender UI. The user can select a specific practitionner from the dropdown list to get a specific schedule from a vet.

In addition, I refactored some methods so that they take less space in the code and deleted duplicates to make it moe efficient.

I removed unhelpful comments as well and fixed some CSS issues related to the implementation of this feature.

### DIFF
--- a/api-gateway/src/main/resources/static/scripts/visits/visits.template.html
+++ b/api-gateway/src/main/resources/static/scripts/visits/visits.template.html
@@ -39,6 +39,7 @@
     </div>
 </div>
 <script>
+
     const date = new Date();
 
     //specify which months goes with which numbers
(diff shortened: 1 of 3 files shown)


api-gateway/src/main/resources/static/scripts/visits/visits.template.html
```diff
@@ -267,29 +313,64 @@ <h1></h1>
...
 
+    //check selection change for vets availability
+    const selectChanged = () => {
+        vetAvailabilityStr = $('#vetWorkdays').val();
+        nameOnCalenderStr = $("#selectedVet :selected").attr("data-target");
+
+        console.log(nameOnCalenderStr);
+
+        renderCalendar();
+
```
Funnyadd, line 363: Have you thought about removing that line afdter the end of the function?

codingbeast666 (approved):
Looks good. Might need to change the fields that will be entered in the dropdowns later but it's good for now.

bthorng (approved):
Beautiful code 😎

PR #128, opened 2021-10-05
feat(VISE-CPC-159): Fetch previous and scheduled visits separately

cjayneb:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-159?atlOrigin=eyJpIjoiZGVkYjViMWExNTk1NGM5MmFhZjMxMzZjZWEzZWY1MTUiLCJwIjoiaiJ9

- Our service now has a GET mapping for fetching previous visits only and a GET mapping for fetching scheduled visits only (for a specific pet).
- Added a method in the business layer to filter out unwanted visits based on a boolean and the date and time when the request is made.
- Added tests for those methods.
- Fetching visits with a negative petId will now throw an InvalidInputException which is handled gracefully.


codingbeast666 (approved):
Good that you also cleanup some of the comments we don't need anymore .

PR #129, opened 2021-10-05
conf(BILL-CPC-242): Add Jacoco to Billing Service

jason-gerard:


partenersintime (approved):
Everything appears to be in order. Jacoco was added to the build.gradle and then tests from the previous sprint were reworked to ensure 90% test coverage. New tests were also added to ensure all aspects of the Billing Service were covered.

billing-service/src/test/java/com/petclinic/billing/presentationlayer/BillResourceTest.java
```diff
@@ -0,0 +1,150 @@
...
+        assertEquals(httpErrorInfo.getHttpStatus(), null);
+        assertEquals(httpErrorInfo.getPath(), null);
+        assertEquals(httpErrorInfo.getTimestamp(), null);
+        assertEquals(httpErrorInfo.getMessage(), null);
+    }
+
+    @Test
+    void testHttpErrorInfoConstructor() {
+        HttpErrorInfo httpErrorInfo = new HttpErrorInfo("timestamp1", HttpStatus.NOT_FOUND, "Bill not found");
+    }
```
cjayneb, line 122: I don't think this test tests anything.

cjayneb: You're right, but I meant that you are not asserting that the values you passed in the constructor are the ones that were set the httpErrorInfo object.

PR #134, opened 2021-10-06
feat(CUST-CPC-288): Added DeletePet in PetResource and created shouldDeletePet persistence test

christopherl04:
Added DeletePet with DeleteMapping in PetResource and created shouldDeletePet persistence test which is way better than the old test.

Please go through the code like Jason showed in class to make sure all is good :)

newTest2

deletemapping


ShiiroFX (approved):
The code is clean and the logic looks good. Also, great job on the comments and keeping the clear style for the title of each section.

PR #135, opened 2021-10-06
feat(VISE-CPC-224):Made the UI to Modify a Visit

lbains:
- Added a button next to the already existing visits that allows us to edit a visit
- Made it so the visit information is filled out in the form and we are allowed to change it
- Added a button that appears that allows us to exit the editing phase 
- Made it so once update visit is clicked, it brings us back to the default form with the existing visits, the edited visit will have its newly modified information 
- Things to note: the VisitDetail class in the API Gateway service does not have the Practitioner Field hence why there are no practitioners shown next to the already existing visits, as well as the date field is by default set to the current date, which is why it is always filled out regardless of being in the edit state or not
Screenshot 2021-10-06 165638
(Editing the visit with the description "Testing Purposes")
Screenshot 2021-10-06 165657
(Once update visit is clicked, the visit is updated and we are presented with the default state of the form)
.


cjayneb (approved):
Great work! Looking forward to testing it locally. LGTM

PR #140, opened 2021-10-07
Feat(VETS-CPC-335): Dockerfiles Error Fix For All Services

Louenes099:
Sorry Christine I was too lazy to commit after each service but the fix is there enjoy everyone!

### DIFF
--- a/billing-service/Dockerfile
+++ b/billing-service/Dockerfile
@@ -1,4 +1,4 @@
-FROM gradle:6 as builder
+FROM gradle:6.9.1 as builder
 WORKDIR /usr/src/app
 COPY src ./src
 COPY build.gradle .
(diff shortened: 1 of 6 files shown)


VladPopa616 (approved):

cjayneb (approved):
LGTM

billing-service/Dockerfile
```diff
@@ -1,12 +1,13 @@
...
+FROM gradle:6.9.0 as builder
 WORKDIR /usr/src/app
 COPY src ./src
 COPY build.gradle .
 RUN ["gradle", "bootJar"]
 
 EXPOSE 8080
 
-FROM openjdk:11
+FROM openjdk:8-alpine
```
Fube, line 9: Are we sure billing wants to use JDK 8?
Maybe they left it as JDK 11 because they rely on JDK 11 specific features?

Louenes099: okay should i change it for eveyone?

Louenes099:
Changing to jdk 8 for the billing service fixed their failing service

PR #143, opened 2021-10-07
feat/(VISE-CPC-213) Make UI for delete visit

NickEspitalier:


bthorng (commented):
Looks good to me, just wondering what the `self.visits.splice(visitId, 1);` does before I approve 👍
image

bthorng (approved):
Looks good to me, had a suspicion that that line of code was redundant 👍

codingbeast666 (approved):
I think making changes to the BFFapigateway files should be done by the APIG team but I will approve because it's a small fix.

PR #144, opened 2021-10-07
feat(CUST-CPC-141):Implemented and created tests for InvalidInputException

christopherl04:
I implemented the invalidInputException in OwnerServiceImpl. I deleted two unnecessary constructors for in the InvalidInputException class because we were not using them and therefore, had no need to test them. Keeping those was impacting our code coverage. I created two tests for the method one for the empty constructor and one for the constructor with a message. Both tests passed and the app would build after my commit so it appears that nothing is broken here. Jacoco tests are also at 100% now for the InvalidInputException.

Please observe and compare to make sure that everything looks good!

CheckingjacocoTestAfterMakingMyTests
BuildSuccessful
firstInvalidInputExceptionPassed
invalidInputexceptionTestPassed


ShiiroFX (approved):
Looks good, great pull request by the way! Description and file usage is perfect for reviewing.

PR #147, opened 2021-10-08
feat(VETS-CPC-237): add vet img field and modified sql files

SmoothWin:
This is a pull request where the vet entity has been modified in order to hold an image field, SQL files were modified accordingly with the new field and some fixes were also added in order to solve an underlying problem.
Here is the diagram displaying the current entity changes:
image
Here are examples of the vet docker image working:
image


amine-12 (dismissed):

Louenes099 (dismissed):
Changed the appropriate file in the appropriate way

amine-12 (dismissed):

amine-12 (dismissed):

amine-12 (approved):

Louenes099 (approved):
Everything is perfect according to me

amine-12 (approved):

PR #148, opened 2021-10-08
conf(CUST-CPC-258): Implemented Swagger UI

ShiiroFX:
Added dependency for Swagger UI in the gradle of the customer service.


Louispoire:
Perfect. Now we have Swagger, which is great!

Anh-Toine (approved):
Nice! This tool will definitely help us with testing out our service.

PR #149, opened 2021-10-09
feat(VISE-CPC-333) : Add sorting by columns for Previous Visits

bthorng:
### Add sorting by columns for Previous Visits
The goal of this story was to add table headers that when clicked would sort the contents of the table based on what was clicked in either ascending or descending order.

**Changes**
* Added some styling on UI to make it look more appealing
* Added table headers that sort the displayed visits based on date, description or vet name
* Modified SQL data to add more visits for pet with the ID 7 to test sorting
* Removed uncessary GET request in visits.controller.js

**Known issues/problems:**
* VisitDetails DTO needs to be updated in order to properly display the vet's name and sort them (Kenny from the API team is handling this)
* Currently, there is only one mapping for Vets service that returns only active veterinarians. This leads to an issue where if a visit had a practitioner that is now inactive, their name would not be displayed in the table. To fix this, we would need to be able to fetch inactive veterinarians from the API. (I believe the vets team already has a mapping in their service for this, they would just need to make on in API)
* Currently, every single vet is fetched and cross-referenced to display their name, if possible in the future this could be changed to request a list of vets based on a list of ID's instead. (while not too much of an issue, performance would be improve if we only had to parse relevant data in the JavaScript)

### Form with no sorting
image

### Sorting visits by descending date
image

### Sorting visits by ascending description
image

### Sorting visits by descending veterinarian name
image


cjayneb (approved):
Great work! LGTM

PR #152, opened 2021-10-09
feat(AUTH-CPC-310): Register user

Fube:
### Changes

- Added encoding logic (using Bcrypt)
- Added spring security rule to allow access to registration endpoint with role or authentication

### DIFF
--- a/auth-service/src/main/java/com/petclinic/auth/User/UserServiceImpl.java
+++ b/auth-service/src/main/java/com/petclinic/auth/User/UserServiceImpl.java
@@ -5,6 +5,7 @@ import org.springframework.dao.DataIntegrityViolationException;
 import org.springframework.dao.DuplicateKeyException;
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.PageRequest;
+import org.springframework.security.crypto.password.PasswordEncoder;
 import org.springframework.stereotype.Service;
 import com.petclinic.auth.Exceptions.NotFoundException;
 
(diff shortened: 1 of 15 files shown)


auth-service/src/main/java/com/petclinic/auth/User/UserServiceImpl.java
```diff
@@ -35,15 +36,10 @@ public User getUserById(long id) {
...
 
-        try {
-            log.info("Saving user with email {}", userIDLessDTO.getEmail());
-            User user = userMapper.idLessDTOToModel(userIDLessDTO);
-            return userRepo.save(user);
-        }
-        catch (DataIntegrityViolationException e){
-            throw new DataIntegrityViolationException("Duplicate email for userEmail " + userIDLessDTO.getEmail());
-        }
-
```
JordanAlbayrak, line 46: I believe this was added by Jeremy for his email validation/verification, I'm wondering if maybe this should have stayed. Next class I'll bring this up to confirm.

JordanAlbayrak (approved):
Thank you for taking the to complete this. It looks good to me, I left a comment not sure if it will show up but I'll ask you in class later anyway.

auth-service/src/main/java/com/petclinic/auth/User/UserServiceImpl.java
Fube: The issue with this is that it assumes that ALL data integrity violations are due to duplicate emails.
This is untrue as a data integrity exception can occur when referencing a non-existent field or using an invalid value (in terms of type) and a myriad of other things.

Trilikin21 (approved):
LGTM

PR #153, opened 2021-10-10
feat(VETS-CPC-233): Upgrade Modal and Table

SmoothWin:
Implemented UI changes for the table and the modal popup on vet name hover.
When the table has all of its data, the modal will only show the vet name and vet specialties as this image displays.
image
When the screen width reaches **768 pixels** the table will no longer display the phone number data it will display the phone number instead of on the modal hover.
image
When the screen width reaches **550 pixels** the table will no longer display the email address data it will be moved to the modal instead.
image
This is generally done in order for the table to support smaller screen sizes without suffering from lower information overall because the values are still present on the page in a way.


SmoothWin:
Proof of both vets service and API gateway service working on docker
image
image

Louenes099 (approved):
All proof and validation are there so it's perfect!

PR #154, opened 2021-10-10
feat(CUST-CPC-142):Implemented and created tests for NotFoundException

christopherl04:
I removed ResourceNotFoundException entirely because we already have a NotFoundException, therefore i changed the few uses of ResourceNotFoundException in the code to use NotFoundException. I created 2 successful tests for NotFoundException and removed the 2 unused constructors in the NotFoundException that was causing a lack of code coverage. Building the app in Git bash was a success, the tests all passed so nothing seems to be broken here. Jacoco test is at 100% for both of our exceptions.

Please observe and compare to make sure that everything looks good!
BothExceptionsFullypassed

AllExceptionsTestsPassing


ShiiroFX (approved):
Great analysis Chris. Tests are our main focus right now and this will help a lot.

PR #155, opened 2021-10-10
feat(VISE-CPC-347): Fetch practioner id by visitId

cjayneb:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-347?atlOrigin=eyJpIjoiMWI1ZTY4NmNlZWJlNGMyZGFlM2ZlZTczZGI4N2FiZTQiLCJwIjoiaiJ9
## Context:
This PR is to add the option to get the practitionerId/vetId that is associated with a specific visit in our database.
## Changes
- I have added a method in our presentation layer with a get mapping to respond to API requests with tests to make sure it handles exceptions properly.
- I have added a method in our service layer that calls the repo to fetch a visit and extracts the practitionerId out of it if no exception is thrown. 
- I have added two tests for the repo method to make sure it is working properly.

## Dev notes (Optional)
- The methods do not cross-reference with the vets database to check if a practitioner/vet with the returned practitionerId actually exists.
- ^^^ We should make sure to use the same format that the vets are using for their public business key.


visits-service/src/test/java/com/petclinic/visits/datalayer/PersistenceTests.java
```diff
@@ -120,5 +122,19 @@ public void updateVisit(){
...
+
+    @Test
+    public void findVisitByIdNonExistentVisit(){
+        Optional<Visit> v = repo.findById(234);
+
+        assertEquals(Optional.empty(), v);
+    }
+
+    @Test
+    public void findVisitByIdValid(){
```
cjayneb, line 134: I dont know why this test doesn't pass with the CI/CD pipeline. The project builds fine when I build it on my end.

PR #160, opened 2021-10-11
bug(MAIL-CPC-359): E-mail body not sending

Fube:
### Changes

- Fixed e-mail body being set as header instead of body


cjayneb (approved):
LGTM

PR #161, opened 2021-10-11
feat(VISE-CPC-222):  Separate Previous and Scheduled Visits UI

bthorng:
### Separate Previous and Scheduled Visits UI

**Context**
The goal of this story was to split and sort visits in the Add Visit Form to include two tables. One table for upcoming visits (sorted based on today's date and onward) and previous visits.

**Changes**
- Added upcoming visits table with same sorting functionality as previous visits
- Sorting a visit into two lists based on date (upcoming & previous visits)
- New stylesheet used by the the form in css/addVisitFormStyle.css
- Added a table row color change on hover
- Changed arrow on sorting buttons
- Modified some database entries for pet with ID 7

**Things to consider**
- Currently, the status field in a Visit does nothing, possibly in the future it could determine whether a visit is in upcoming visits or previous visits

### Add Visit Form displaying upcoming visits and previous visits
image

### DIFF
--- a/api-gateway/src/main/resources/static/scripts/visits/visits.controller.js
+++ b/api-gateway/src/main/resources/static/scripts/visits/visits.controller.js
@@ -21,14 +21,15 @@ angular.module('visits')
             var dd = String(dateObj.getDate()).padStart(2, '0');
             var mm = String(dateObj.getMonth() + 1).padStart(2, '0');
             var yyyy = dateObj.getFullYear();
-            let currentDate = Date.parse(yyyy + '/' + mm + '/' + dd);
+            let currentDate = Date.parse(yyyy + '-' + mm + '-' + dd);
 
             self.upcomingVisits = [];
             self.previousVisits = [];
 
             $.each(self.visits, function(i, visit) {
                 let selectedVisitDate = Date.parse(visit.date);
-                if(selectedVisitDate > currentDate) {
+
+                if(selectedVisitDate >= currentDate) {
                     self.upcomingVisits.push(visit);
                 } else {
                     self.previousVisits.push(visit);


api-gateway/src/main/resources/static/scripts/visits/visits.controller.js
```diff
@@ -13,8 +13,30 @@ angular.module('visits')
...
+                let selectedVisitDate = Date.parse(visit.date);
+
+                if(selectedVisitDate >= currentDate) {
+                    self.upcomingVisits.push(visit);
+                } else {
+                    self.previousVisits.push(visit);
+                }
+            });
+        }
+
```
cjayneb, line 39: The point of my back-end story (fetch previous and scheduled visits separately) was so that you didn't have to do this logic in the front-end.

cjayneb (approved):
Good work, although there is some logic for sorting that could've been passed on to our service since I made methods specifically to sort previous and scheduled visits. However, doing it in the front-end reduces the amount of requests needed so maybe it is better.

api-gateway/src/main/resources/static/scripts/visits/visits.controller.js
bthorng: For sure, I'll refactor the code to use our service's methods once we get the mappings for them in API.
It should be better to do it this way as well since there's no sorting needing to be done and the visits can be displayed ASAP, I believe.

cjayneb: Ohhhh you're right I forgot that we had to go through the API

lbains (approved):
Nice work, hopefully we can get the API mappings so that we could use our service's methods.

PR #163, opened 2021-10-11
conf(VISE-CPC-362): Changed the colour of the edit button

lbains:
**Changes**
The edit button looked like a field before, so to make it stand out I changed the colour of the button to blue.

**Why Blue?**
The mapping for edit is a Put Mapping and the Postman colour for a Put request is blue and blue is nice.

**This is how the edit button looked like before.**
Screenshot 2021-10-11 182413

**This is how the edit button looks like now.**
Screenshot 2021-10-11 181959


cjayneb (approved):
Very cool! Blue is definitely the way to go. Lgtm

PR #171, opened 2021-10-12
feat(VETS-CPC-228): add dto and vet mapper

BunTymofiy:
image

Usage of vetDTO is proven by absence of "id"


Louenes099 (approved):
All proof are there, codes looks perfect and no errors so everything is perfect

Iana-Sha (approved):
Looks awesome and has evidence that it work. Good job!

PR #174, opened 2021-10-12
bug(CUST-CPC-367): Fixing Owner Table

CrxJr:
Fixed the table not been created for Customer Service

### DIFF
--- a/api-gateway/src/main/resources/static/scripts/owner-details/owner-details.controller.js
+++ b/api-gateway/src/main/resources/static/scripts/owner-details/owner-details.controller.js
@@ -8,11 +8,11 @@ angular.module('ownerDetails')
             self.owner = resp.data;
         });
 
-        $http.update('api/gateway/owners/' + $stateParams.ownerId).then(function (resp) {
+        /*$http.update('api/gateway/owners/' + $stateParams.ownerId).then(function (resp) {
             self.owner = resp.data;
         });
 
         $http.delete('api/gateway/owners/' + $stateParams.ownerId).then(function (resp) {
             self.owner = resp.data;
-        });
+        });*/
     }]);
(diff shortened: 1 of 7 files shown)


api-gateway/src/main/resources/static/scripts/owner-details/owner-details.controller.js
```diff
@@ -8,11 +8,11 @@ angular.module('ownerDetails')
...
 
-        $http.update('api/gateway/owners/' + $stateParams.ownerId).then(function (resp) {
+        /*$http.update('api/gateway/owners/' + $stateParams.ownerId).then(function (resp) {
             self.owner = resp.data;
         });
 
         $http.delete('api/gateway/owners/' + $stateParams.ownerId).then(function (resp) {
             self.owner = resp.data;
-        });
+        });*/
```
Louispoire, line 17: I'm not sure I understand why these lines might be the ones causing the issue. We need to make sure that we still keep in order to call an update/delete HTTP request!

customers-service/src/main/resources/data-h2.sql
```diff
@@ -0,0 +1,31 @@
+INSERT INTO types VALUES (1, 'cat');
```
Louispoire, line 1: Good thing you added theses files back

api-gateway/src/main/resources/static/scripts/owner-details/owner-details.controller.js
```diff
@@ -8,11 +8,11 @@ angular.module('ownerDetails')
             self.owner = resp.data;
         });
 
-        $http.update('api/gateway/owners/' + $stateParams.ownerId).then(function (resp) {
+        /*$http.update('api/gateway/owners/' + $stateParams.ownerId).then(function (resp) {
```
Louispoire, line 11: We should create 2 new JS file, each one containing an HTTP request like update and delete

Louispoire (approved):
Good stuff. By looking at your code, I see there was some issue with the angular files and html. Glad that you also re-added the sql files

Anh-Toine (approved):
Looks great! We should be able to do curl requests now

PR #178, opened 2021-10-13
feat(CPC-134, CPC-132, CPC-309): Create/get/delete/update a single vet

tristangiguere:


VladPopa616 (approved):
Overall seems ok, but each story should have been a different pull request and watch your test naming. Name them something like "shouldCreateAndDeleteVets". I will still approve it though. One more thing, you forgot the team tag inside the pull request name.

Ludo67 (approved):
have you considered changing your test names ? Also, different methods should have their own PRs

Louenes099:
FYI I was being respectful as much as I could because i'm a PO for this sprint. You don't know me and you don't wanna know me. College was peaceful for me until now and I still want it to be so don't push it.

You poked the wrong bear at the wrong time. I told you it's a warning. Until now. Last Chance. Even Christine or Istiaque won't be able to stop me if you push it. Just be happy I didn't delete all your stuff.

PR #179, opened 2021-10-15
feat(AUTH-CPC-388): Add login

Fube:
### Changes

- Added log in functionality (JWT generation)
- Added Spring Security configuration for JWT filter
- Added E2E tests for register, verify, login, and accessing protected routes (i.e. /roles)
- Added exception handling for bad login
- Fixed MapStruct setting roles to `null`


auth-service/src/test/java/com/petclinic/auth/User/AuthServiceUserEntityTests.java
```diff
@@ -76,4 +86,60 @@ void user_builder() {
...
+    void isAccountNonLocked() {
+        assertTrue(new User().isAccountNonLocked());
+    }
+
+    @Test
+    @DisplayName("Given any user, credentials is non expired")
+    void isCredentialsNonExpired() {
+        assertTrue(new User().isAccountNonExpired());
+    }
+
```
cjayneb, line 133: couldn't you combine these into a single test?

Fube: I could but if in the future they branch out to have real algorithms and implementations (instead of `return true`), it would be more work

cjayneb (approved):
LGTM

PR #180, opened 2021-10-16
feat(APIG-CPC-354): Integrate register

Fube:
### Changes

- Added exception handler to gateway
- Fixed create user invoking wrong URL
- Fixed create user not sending body
- Fixed create user not calling correct service method
- Added Rethrower helper class for my own sanity (fuck Mono)
- Added GenericHttpException & HttpErrorInfo
- Added Register DTO
- Added missing auth client service tests

#### Angular Stuff

- Fixed sign up controller making a useless get request
- Fixed sign up controller's `add` not being called
- Fixed sign up controller's `$scope.signup` not aligning with the template (signup VS newUser)
- Added (soft) redirect to login page after successful register
- Added error display on unsuccessful register

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -68,7 +68,8 @@ services:
     environment:
       - SMTP_USER=champlain.petclinic@gmail.com
       - SMTP_SERVER=smtp.gmail.com
-      - SMTP_PASS=nicetryguy
+    env_file:
+      - mailer.env
 
 
   mysql1:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -46,7 +46,9 @@ services:
       - "7005:8080"
     depends_on:
       mysql4:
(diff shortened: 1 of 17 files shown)


docker-compose.yml
```diff
@@ -68,7 +68,8 @@ services:
     environment:
       - SMTP_USER=champlain.petclinic@gmail.com
       - SMTP_SERVER=smtp.gmail.com
-      - SMTP_PASS=nicetryguy
```
JordanAlbayrak, line 71: not sure exactly why we would not have a pass so I'll leave a comment here and maybe receive some insight, not like I really fully understand what's going on here so that insight my just go through me.

Fube: I removed it because it (should) have higher priority than the .env file where the real `SMTP_PASS` is loaded from

PR #182, opened 2021-10-16
feat(AUTH-CPC-460): Exception handling

Fube:
### Changes

- Added exception handling for bad password on register
- Added exception handling for dupe email

#### Invalid e-mail and password
image

#### Invalid password
image

### Dupe email
image


JordanAlbayrak (dismissed):
LG2M, Max, Jeremy and I spent some time on it yesterday but we couldn't figure it out. Thanks for being big brained and fixing things :))

Trilikin21 (approved):
After carefully reviewing your additions. Its looking good to me.

auth-service/build.gradle
```diff
@@ -82,6 +82,7 @@ jacocoTestReport {
 		classDirectories.setFrom(files(classDirectories.files.collect {
 			fileTree(dir: it, exclude: [
 					"com/petclinic/**/AuthServiceApplication.class",
+					"**/*MapperImpl.class"
```
cjayneb, line 85: Why is this class excluded from the jacoco report? Is it because it's automatically generated?

auth-service/src/main/java/com/petclinic/auth/User/UserController.java
```diff
@@ -65,7 +67,9 @@ public User getUser(@PathVariable long userId) {
     }
 
     @PostMapping
-    public UserPasswordLessDTO createUser(@RequestBody @Valid UserIDLessRoleLessDTO dto) {
+    public UserPasswordLessDTO createUser(
+            @RequestBody @Valid UserIDLessRoleLessDTO dto,
+            BindingResult bindingResult) {
```
cjayneb, line 72: Where is this bindingResult used?

auth-service/src/test/java/com/petclinic/auth/e2e/User/AuthServiceE2ETests.java
```diff
@@ -190,10 +193,59 @@ void admin_access_protected_route() throws Exception {
...
+            put("password", USER.getPassword());
+        }});
+
+        final MvcResult result = mockMvc.perform(post("/users/login").contentType(APPLICATION_JSON).content(asString))
+                .andExpect(status().is4xxClientError())
+                .andExpect(content().contentType(APPLICATION_JSON))
+                .andExpect(jsonPath("$.password").doesNotExist())
+                .andExpect(jsonPath("$.statusCode").value(UNAUTHORIZED.value()))
+                .andExpect(jsonPath("$.timestamp").exists())
+                .andExpect(jsonPath("$.message").value(format("Password not valid for email %s", USER.getEmail())))
```
cjayneb, line 226: Shouldn't this expect a message saying that there is no user registered with the given email?

auth-service/build.gradle
Fube: Yes.
I still wrote relevant mapper tests.
From my research, neither JaCoCo nor MapStruct expose a preferable solution to this

auth-service/src/main/java/com/petclinic/auth/User/UserController.java
Fube: It is not nor should it be.
For some ungodly reason, Java X Validation does NOT throw the appropriate exception if you do not take in the `BindingResult` in your method params.

PR #183, opened 2021-10-17
feat(APIG-CPC-461): Integrate email verification

Fube:
### Changes

- Integrated email verification

#### Expired token
image

#### Bad token
image

#### Really bad token
image


VladPopa616 (approved):
Didn't really have the time to review this earlier, but from what I am seeing, I think it looks good. Maybe just explain to me how does the token work in the URL path.

Fube:
> Maybe just explain to me how does the token work in the URL path.

It's a base64(JWT) and its subject is set to email that needs to be confirmed
You give that to auth who verifies the JWT signature, and if it is correct, it let's go through to the service which just does `model.verified = true`

VladPopa616 (approved):
So I guess my review got dismissed because of merge conflicts, but basically, yeah, same comment as my previous one.

PR #187, opened 2021-10-18
bug(VETS-CPC-468): field naming fix for mapping

amine-12:
JIRA: https://champlainsaintlambert.atlassian.net/jira/software/c/projects/CPC/boards/13?modal=detail&selectedIssue=CPC-468&sprint=72
## Context:
This is a required fix because the class cannot be mapped correctly if the field names are different from the vet-service field names
## Changes
I changed private Integer enabled; to private Integer isActive;


Louenes099 (approved):
Made the correction I asked and all the requirements are present so hence I consider this feature as done with the power that was given to me as a PO.

PR #191, opened 2021-10-19
feat(APIG-CPC-467): Integrate login

Fube:
JIRA: [CPC-467](https://champlainsaintlambert.atlassian.net/browse/CPC-467)

## Context:
Gotta integrate login so that the user can use the application.
Also had some visual aspects to fix (i.e. register button on login)

## Changes
- Added AngularJS global http 401 & 403 handling
- Added authProvider
- Added conditional rendering of navbar (i.e. when logged in, don't show register and login buttons)
- Added integration (proxying) for APIG x AUTH
- Fixed login and register not having same look and feel
- Centered login and register button (on their respective pages)

##### Successful 
image
image

##### Unsuccessful
image

#### Old Login page 😞
image

## Dev notes
- Left some notes for the next maintainer, may God have mercy on his soul
- Heavy limitations imposed by ReactiveWeb making it so that I cannot return meaningful error messages. Forced to just default to "Unauthorized". Needs review at a later point
- **!!IMPORTANT!!** When this PR is merged, all Selenium tests will require the user to register AND confirm their email THEN login to be able to access anything in the app. I am planning on adding a simple in-mem auth provider to fix that. A story will be made to address this.

### DIFF
--- a/api-gateway/src/main/java/com/petclinic/bffapigateway/domainclientlayer/VisitsServiceClient.java
+++ b/api-gateway/src/main/java/com/petclinic/bffapigateway/domainclientlayer/VisitsServiceClient.java
@@ -12,6 +12,7 @@ import org.springframework.web.reactive.function.client.WebClient;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
+import java.util.Date;
 import java.util.List;
 
 import static java.util.stream.Collectors.joining;
@@ -54,8 +55,6 @@ public class VisitsServiceClient {
                 .bodyToMono(Visits.class);
     }
 
-
-
     public Flux<VisitDetails> getVisitsForPet(final int petId){
         return webClientBuilder.build()
                 .get()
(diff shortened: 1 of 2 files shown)


Kenny-star (approved):
Everything looks good to me, but is it possible to provide an explanation for the Login method in the authserviceclient? I'm a mesmerized by the code you wrote. :)

api-gateway/src/main/resources/static/scripts/app.js
```diff
@@ -1,10 +1,72 @@
...
+            token: $window.localStorage.getItem("token"),
+            username: $window.localStorage.getItem("username"),
+            email: $window.localStorage.getItem("email"),
+        }),
+        purgeUser: () => {
+            $window.localStorage.removeItem("token")
+            $window.localStorage.removeItem("username")
+            $window.localStorage.removeItem("email")
+        },
+        isLoggedIn: () => !!$window.localStorage.getItem("token")
```
cjayneb, line 32: Not sure I get what !! does here

cjayneb: Im guessing it converts the value to a boolean? but how

cjayneb (approved):
Not sure I understand everything fully, but it makes sense from what I gathered. Should I base myself on this if I ever want to modify the layout of the UI based on the login?

api-gateway/src/main/resources/static/scripts/app.js
Fube: [Because of how truthy/falsy works](https://developer.mozilla.org/en-US/docs/Glossary/Truthy)

PR #193, opened 2021-10-19
feat(VISE-CPC-384): Cancel a Visit Through UI 

lbains:
JIRA: [CPC-384](https://champlainsaintlambert.atlassian.net/browse/CPC-384?atlOrigin=eyJpIjoiZmM2MWFiN2FjN2FjNDkyZThlYWZkNzBiMWVjZTM1MWUiLCJwIjoiaiJ9)
## Context:
If a customer needs to cancel a visit, an employee will need to cancel the visit through the UI. We do not want to completely delete this visit from the database, but instead indicate that it has been canceled. This is why a status field was added earlier in Sprint 1. Now with these changes, there is a way to cancel a visit through the UI. This will change the status of the visit and will display it next to the visits that were fetched. If for some reason, the cancel needs to be undone, the button may be clicked once more to undo the cancel.
## Changes
- Added a new button that will actually cancel the visit
- Added a new column in the table that will show the status of the visit
- Added sorting by the status as well
## Before and After UI 
**Before (Without the new Cancel Button and Status Column)**
Screenshot 2021-10-19 164322
**After(UI with the new changes)**
Screenshot 2021-10-19 164542


api-gateway/src/main/resources/static/scripts/visits/visits.controller.js
```diff
@@ -267,11 +269,60 @@ angular.module('visits')
...
+                    self.previousVisits.sort(function (a, b) {
+                        a = self.getStatus(a.status).toLowerCase();
+                        b = self.getStatus(b.status).toLowerCase();
+
+                        return a > b ? -1 : a < b ? 1 : 0;
+                    });
+                    $('#sortByStatusButtonPreviousVisits').text("Sort by status ↑")
+                }
+            }
+        }
```
cjayneb, line 318: There must be a way to call the same functions for sorting the previous visits table and scheduled visits table instead of calling a separate function every time.

cjayneb (approved):
Other than my comment about redundancy. Looks good to me!

api-gateway/src/main/resources/static/scripts/visits/visits.controller.js
lbains: For both previous and scheduled, it will call the SortTableByStatus function, like Brandon mentioned, the boolean parameter decides whether it is previous or scheduled.

api-gateway/src/main/resources/static/scripts/visits/visits.controller.js
```diff
@@ -294,4 +345,66 @@ angular.module('visits')
...
+                    practitionerId: visitPractitionerId,
+                    status: true
+                };
+            }
+
+
+            url = "api/gateway/owners/*/pets/" + petId + "/visits/" + visitId;
+
+            $http.put(url, data).then(function () {
+                $http.get("api/gateway/visits/"+petId).then(function (resp) {
```
bthorng, line 386: We probably shouldn't be requesting all the visits again when we're just changing the status of a visit. Maybe request that single visit and then update the out of date visit in the list of visits that was already fetched to match it?

api-gateway/src/main/resources/static/scripts/visits/visits.controller.js
```diff
@@ -294,4 +345,66 @@ angular.module('visits')
...
+                    self.visits = resp.data;
+                    self.sortFetchedVisits();
+                });
```
bthorng, line 389: We probably shouldn't be requesting all the visits again when we're just changing the status of a visit. Maybe request that single visit and then update the out of date visit in the list of visits that was already fetched to match it?

bthorng (approved):
We don't have a way to get a single visit in our service so I'll allow it 🤠

PR #194, opened 2021-10-20
bug(BILL-CPC-449): Configure logic for a not found visit type

LemireCoding:
JIRA: https://champlainsaintlambert.atlassian.net/jira/software/c/projects/CPC/boards/15?modal=detail&selectedIssue=CPC-449&assignee=6005c6881ffcc00076745e8b
## Context:
Configure logic for a not found visit type. The code for the visit type to price was done in sprint 2 but I realized we were missing logic for when a visit type is not found in our list of possible visit type
## Changes
Changed the tests and added to the CreateBill method of Billing service which now throws and exception if the get to the hashmap returns null


Trim0500 (approved):
I'll approve this. The code makes sense and is readable. It is coupled with a test and the acceptance criteria is met. Nice work!

Dratorz (approved):
Code is well written and the assert is used correctly. Very good.

PR #195, opened 2021-10-20
feat(AUTH-CPC-486): Default admin account

Fube:
JIRA: [CPC-486](https://champlainsaintlambert.atlassian.net/browse/CPC-486)
## Context:
Gotta have a default login with role ADMIN so that someone can assign that role to others.

## Changes
- Added in-memory authentication provider which holds only one user, the default admin
- Default username and password for the user can be changed by modifying `default-admin.username` and `default-admin.password` respectively

## Linked pull requests
- [Integrate Login](https://github.com/cgerard321/champlain_petclinic/pull/191)


Fube:
> Yes it should be good, if I understand correctly, the default username and password would be admin and it would give everyone access to the website

You are correct yes

NickEspitalier (approved):

MaxGrabs (approved):
looks even better

PR #200, opened 2021-10-20
bug (AUTH-CPC-489): clear text pwd that works in all browsers

ArianneBonneau:
JIRA: https://champlainsaintlambert.atlassian.net/jira/software/c/projects/CPC/boards/16?modal=detail&selectedIssue=CPC-489&quickFilter=29
- fixed the clear text password to make it work in all browsers
- only added a <i> tag to the UI
- Used javascript onclick function

CHROME:
Screenshot (203)
EDGE:
Screenshot (204)


MaxGrabs (approved):
lgtm

PR #201, opened 2021-10-21
Feat/VISE-CPC-111 Add Visit Type

NickEspitalier:
JIRA: link to jira ticket
https://champlainsaintlambert.atlassian.net/jira/software/c/projects/CPC/boards/12/backlog?view=detail&selectedIssue=CPC-111&issueLimit=100 

It is a work a progress I think so guys from the visits and bills teams, listen up, we need to find a way to send the visit type from the visits services to the bills services. Thing is the only way to do that would be by creating a get or a post mapping somewhere but that's where your review will be helpful to find out what to do exactly. Please be precise. 

As for what works, the design for the label have been modified and the select boxes work.

old design
new design

### DIFF
--- a/api-gateway/src/main/java/com/petclinic/bffapigateway/domainclientlayer/BillServiceClient.java
+++ b/api-gateway/src/main/java/com/petclinic/bffapigateway/domainclientlayer/BillServiceClient.java
@@ -6,6 +6,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 import org.springframework.web.reactive.function.client.WebClient;
+import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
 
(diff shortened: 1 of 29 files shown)


bthorng (commented):
I don't like the new formatting, I feel like it creates wasted space, we should keep it vertical.
Also, the visit type selection label should just say `Visit Type:`

NickEspitalier:
Brandon, the design as been changed to what it was originally, as for the functionality of the post method Jean-Christophe I am currently developing this part. Let's hope I finish this feature soon for the bill team

cjayneb:
> new design 2
> 

I think @bthorng wanted the visit type select to be under the practitioner information because it looks a bit out of place when it's next to the practitioner select

bthorng:
> 
> 
> > new design 2
> 
> I think @bthorng wanted the visit type select to be under the practitioner information because it looks a bit out of place when it's next to the practitioner select

Wait, that's actually a good point though. I think the form would look better that way.

NickEspitalier:
final design

PR #220, opened 2021-10-23
FEAT(APIG-CPC-501) Create test_all.bash

LawrenceLiang02:
Added the test_all.bash file and added the host and port. It is ready to be used once the aggregates are created.

JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-501?atlOrigin=eyJpIjoiMTBjYmJlOThiNWIyNDgyMjg1ZjA5ZTEyZDc3OTM1NTUiLCJwIjoiaiJ9
## Changes
Added a file called test_all.bash so that my teammates can write their tests once they are done with the aggregate class.


VladPopa616 (approved):
Perfect! Can't see why I wouldn't approve this.

PR #283, opened 2022-09-29
Feat/bill cpc 537 upgrade billing micro service to be reactive

C-hunker:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-537
## Context:
All previous code in the BillingService is now reactive including the tests.
### Changes
- Changed BillingService repository to feature a reactive mongo repository.
- Implemented new reactive methods for the new database.
- Implemented reactive testing for all the new reactive methods implemented.
- Changed Jacoco version to 0.8.8
## Dev notes 
- Currently exceptions are not implemented in the BillingService. The two classes that are not tested are BillControllerExceptionHandler and HttpErrorInfo. Jacoco currently excludes those files when running tests.
- Jacoco coverage reaches 98% in this current state.
- BillResourceUnitTest has an issue where the insert test is not functioning. It should be fixed by the next sprint.


Marina1411 (approved):

billing-service/src/test/java/com/petclinic/billing/exceptions/InvalidInputExceptionTest.java
```diff
@@ -30,4 +30,12 @@ void TestInvalidInputMessageAndCauseConstructor() {
             throw new InvalidInputException("Something went wrong", new Exception());
         });
     }
+
+
+    @Test
+    void TestInvalidInputMessageOnly() {
+        assertThrows(InvalidInputException.class, () -> {
+            throw new InvalidInputException("Something went wrong");
```
ElricVermaut, line 38: Both input exception and not found exception return the same error message "Something went wrong", do you think we should give specific messages to each exception

ElricVermaut (approved):
The structure of the service seems strong, there are a lot of things commented that im not sure we will use in the future.

as a sidenote, this pull request was very large (about 800 lines added), which made it hard to find specific things that could be improved. Perhaps reduce the future pull requests to about 300 lines added max. This will help give better feedback

PR #288, opened 2022-10-05
(bug/VVS-CPC-634) fix docker conflict

Kam1st:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-634
## Context:
This ticket is because we had a conflict in the docker-compose.yml and so we're creating 2 different hosts for the billing and vet services
## Changes
- changed docker-compose.yml to include 2 mongo hosts
- changed the ports to run on different ports

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -78,7 +78,7 @@ services:
         environment:
             - SPRING_PROFILES_ACTIVE=docker
         depends_on:
-            - mongo
+            - mongo1
 #            mysql5:
 #                condition: service_healthy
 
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -78,7 +78,7 @@ services:
         environment:
             - SPRING_PROFILES_ACTIVE=docker
         depends_on:
-            - mongo
+            - mongo1
 #            mysql5:
```
cubajosh, line 82: I would review the commented code because we don't need them

cubajosh (approved):

swafit (approved):
This would all appear to be proper code. 

[nitpick] Perhaps adding a screenshot showing the data being outputted on the end product, to demonstrate that billing and vet databases are functioning properly would help prove this fix has worked.

PR #296, opened 2022-10-07
feat(ACMS-CPC-628): Fix User Validation

DmitryGrigoryev21:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-628?atlOrigin=eyJpIjoiZWQ3NGI3YWVmMmZiNGExZGJkYWJlNDg0OGU3MTU2YWYiLCJwIjoiaiJ9

## Context:
The registration of new users was impossible as the Validation of a new user email would always fail.
Now, the email sent to the registering user will be correct and call the right gateway method to actually validate the user.
The user will now be able to sign in!

## Changes
Api Gateway origin was adjusted to the correct one.

No UI changes where made, but you can now log in and it looks like so:
Screen Shot 2022-10-07 at 7 16 50 PM

### DIFF
--- a/auth-service/src/main/resources/application.yml
+++ b/auth-service/src/main/resources/application.yml
@@ -37,7 +37,7 @@ spring:
   flyway:
     url: jdbc:mysql://mysql4
 gateway:
-  origin: localhost:8080
+  origin: localhost:8080/api/gateway
   protocol: http
   subdomain:
 mailer-service:


felixlabrie (approved):
Tested

AmirHamzahMirza (approved):
Looks good, see no issues with it

PR #314, opened 2022-10-13
bug/ACMS-CPC-686_OwnerPersistenceupdate_owner_test not saving Id correctly

DmitryGrigoryev21:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-686?atlOrigin=eyJpIjoiZWQ4Y2Q1OWViMzRiNGFlMjhmNmJkYmQxMzExNWI5YTMiLCJwIjoiaiJ9

## Context:
update test had a chance to fail depending in what order tests ran since it would save with the wrong id

## Changes
I resigned the generated id on save to the local object to which we are comparing the saved one


felixlabrie (approved):
fixed the problem with IDs not matching

StaceyJin (approved):
Changes applied look sound, programming appears functional.

PR #324, opened 2022-10-16
feat(VVS-CPC-639): Fix Filtering For Active/Inactive in Vet List

Kam1st:
[JIRA: link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-639)
## Context:
we made this ticket to fix the filtering problem in the API to show active/inactive
## Changes
- made changes to the controller to use the endpoints in the API controller instead of the filtering method
- changes the HTML of the dropdown to reflect the new code


Brandon84-alt (approved):
Good!

cubajosh (approved):

PR #340, opened 2022-10-23
Feat/acms cpc 684 rework css for owner details

AmirHamzahMirza:
… Details page

Moved images to center locations right beside tables to be more appealing to view, some other minor changes to clean up the page was done as well. OwnerAggregate has no changes, was only edited to make sure the build went through correctly.

JIRA: 
https://champlainsaintlambert.atlassian.net/browse/CPC-684?atlOrigin=eyJpIjoiNmE4OTk2YzlmZTA2NGYxZDg4YWQwYWM0ZmE0MDUwNzciLCJwIjoiaiJ9

## Context:
Images weren't centered nicely, were moved to fix this. 

## Changes
Moved image and input for image to the right to be more centered in the page, both for pets and owners.
Hover display removed since it does not display correctly due to the already existing design. 

Centered image with image inserted to check if other not visible aspects are changed as well 
image


felixlabrie (dismissed):
tested and working

PR #348, opened 2022-10-23
(Feat/VVS-CPC)695_insert_photo

Brandon84-alt:
JIRA: link to jira ticket
https://champlainsaintlambert.atlassian.net/browse/CPC-695
-Enables Vet to Uplaod Photo on Click/Hover on Photo Icon.
-Should Upload a photo
-Default Photo should appear on hover in Vets Form
 
-When you over on Vet Form you can see vet Speciatlty and Default photo (null for the moment)

Dev notes (Optional)
Photo Fix to assign to VetId is coming in branch 694_Fix_Photo_for_Vet


api-gateway/src/main/resources/static/scripts/vet-details/vet-details.controller.js
```diff
@@ -3,6 +3,7 @@
 angular.module('vetDetails')
     .controller('VetDetailsController', ['$http', '$stateParams', function ($http, $stateParams) {
         var self = this;
+        //var vetId = $stateParams.vetId || 0;
```
Kam1st, line 6: Let's not forget to take out the commented code we have, just to make it clean.

vet-service/src/main/java/com/petclinic/vet/dataaccesslayer/Photo.java
```diff
@@ -0,0 +1,20 @@
+//package com.petclinic.vet.dataaccesslayer;
```
Kam1st, line 1: Is this class necessary?

Kam1st (approved):
Going to have to check out the commented code to see what there is to keep/get rid of.

vet-service/src/main/java/com/petclinic/vet/dataaccesslayer/Photo.java
Brandon84-alt: For my second story (694) which is connected to this user storie will need this package for photos implementation in Backend.

api-gateway/src/main/resources/static/scripts/vet-details/vet-details.controller.js
Brandon84-alt: Will also be needed to have a vetId variable declare for user storie (694) on Joira

cubajosh (approved):

PR #350, opened 2022-10-24
feat(VVS-CPC-639): Implement vets update method in UI

Kam1st:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-692
## Context:
This was to properly implement the update method in the UI because it was not showing all the fields properly. 
## Changes
- Changed the controller.js for the form to display defaults for each value
- Had to change the downstream for naming conventions due to booleans being changed automatically in Spring Boot


cubajosh (approved):

Brandon84-alt (approved):
Naming Convention makes more sense then the old code, good!

PR #367, opened 2022-10-25
Bug/api cpc 707 broken due to improper implementation with the update making the customer micro service reactive

swafit:
JIRA: link to jira ticket
## Context:
## Changes
## Dev notes (Optional)
- pull request link


swafit (commented):
review for self-verification purposes

PR #386, opened 2023-09-16
feat(Cpc 757 invt ability to create inventories)

SandwichKing978:
JIRA: link to jira ticket 
https://champlainsaintlambert.atlassian.net/browse/CPC-757
Context:
In this tickets, I created two features such as creating an inventory and updating an inventory by its Id.
Changes
-update  inventory method added inside Inventory service client
create inventory method added inside Inventory service client
I added both the update and create inventory method inside the bffapigatewaycontroller class.
My changes would affect the Inventories.
It affects the inventory entity


AlexeiTimbro (approved):
The code looks good I approve of these amazing changes.

PR #390, opened 2023-09-17
bug(VIST-CPC-759): fixed visits repository to use String data type as document identifier

swafit:
JIRA: [CPC-759](https://champlainsaintlambert.atlassian.net/browse/CPC-759)
## Context:
A String document type should be used instead of Integer.
## Changes
Changed the visits repo type to String.
No UI change.
## Dev notes
This was separate in its own pull request since multiple people rely on this change for the database to work properly.


CptDoritos (approved):
This passes the vibe check

GiulianoD01 (approved):
looks good

PR #396, opened 2023-09-18
Feat/auth cpc 768 auth force login to access endpoints in api gateway

DylanBrass:
JIRA: [link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-768?atlOrigin=eyJpIjoiM2JjNzU4N2VkZThiNDBjYWJmNzVmODIyNzcxNzQ0MTQiLCJwIjoiaiJ9)
## Context:
This secures the api-gateway. Now every request will need authentication if not annotated with @SecuredEndpoint(allowedRoles = {Roles.ANONYMOUS}).
## Changes
-Added 2 filters one for the token and one for the roles.
-Added exception handling in servlet.
-Made front end work with auth.
image

image

It already existed but now it works !

## Dev notes (Optional)
-The @SecuredEnpoint is created by me so any question about it should be asked to me.

### DIFF
--- a/api-gateway/src/main/java/com/petclinic/bffapigateway/config/GlobalExceptionHandler.java
+++ b/api-gateway/src/main/java/com/petclinic/bffapigateway/config/GlobalExceptionHandler.java
@@ -1,8 +1,7 @@
 package com.petclinic.bffapigateway.config;
 
-import com.petclinic.bffapigateway.exceptions.ExistingVetNotFoundException;
-import com.petclinic.bffapigateway.exceptions.GenericHttpException;
-import com.petclinic.bffapigateway.exceptions.HttpErrorInfo;
+import com.petclinic.bffapigateway.exceptions.*;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.bind.annotation.RestControllerAdvice;
@@ -16,6 +15,7 @@ import java.security.GeneralSecurityException;
  * Date: 2021-10-15
  * Ticket: feat(APIG-CPC-354)
  */
+
 
 @RestControllerAdvice
 public class GlobalExceptionHandler {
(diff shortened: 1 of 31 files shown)


api-gateway/src/main/java/com/petclinic/bffapigateway/config/GlobalServletExceptionHandler.java
```diff
@@ -0,0 +1,61 @@
...
+
+@Slf4j
+@Component
+public class GlobalServletExceptionHandler implements ErrorWebExceptionHandler {
+
+    @SuppressWarnings("NullableProblems")
+    @Override
+    public Mono<Void> handle(ServerWebExchange exchange, Throwable ex) {
+        HttpStatus status = HttpStatus.UNPROCESSABLE_ENTITY;
+
```
RichardDanon, line 23: Is it a good idea to set all your unknown exceptions that have not been specified to HttpStatus.UNPROCESSABLE_ENTITY ? Well, maybe it might not matter in the end since more exceptions will be added.

auth-service/src/main/java/com/auth/authservice/security/JwtTokenUtil.java
```diff
@@ -72,20 +77,16 @@ public String generateToken(User user) {
         Map<String, Object> claims = new HashMap<>();
         claims.put(CLAIM_KEY_USERNAME, user.getUsername());
         claims.put(CLAIM_KEY_CREATED, new Date());
+        ArrayList<String> roles = new ArrayList<>();
```
RichardDanon, line 80: You can directly convert the roles to an array if you want Or use a stream to improve the code maybe

auth-service/src/test/java/com/auth/authservice/presentationlayer/User/UserControllerIntegrationTest.java
```diff
@@ -0,0 +1,104 @@
...
+                .bodyValue(token)
+                .exchange()
+                .expectStatus().isOk();
+
+
+
+    }
+
+
+
```
RichardDanon, line 62: This is for the entire test class but your spacing is kind crazy here try not to add too many unnecessary spaces

RichardDanon (approved):
Well, nothing in here will stop the code from running, only a few changes that you could do to increase the accuracy or efficiency of the code.

drallo22 (dismissed):
Everything looks great. Doesn't seem to conflict with any of our Billing code our cause any previous code to malfunction. Testing all looks very good. Very nice!

auth-service/src/main/java/com/auth/authservice/security/JwtTokenUtil.java
RichardDanon: something like this :

claims.put(CLAIM_KEY_ROLES, user.getRoles().stream().map(role -> role.getName()).toArray(String[]::new));

drallo22 (approved):

api-gateway/src/main/java/com/petclinic/bffapigateway/config/GlobalExceptionHandler.java
```diff
@@ -32,4 +34,37 @@ public ResponseEntity<HttpErrorInfo> resourceNotFoundException(ExistingVetNotFou
...
+                .body(new HttpErrorInfo(498, ex.getMessage()));
+    }
+
+
+    @ExceptionHandler(value = RuntimeException.class)
+    public ResponseEntity<HttpErrorInfo> runtimeException(RuntimeException ex) {
+
+        return ResponseEntity.status(HttpStatus.UNPROCESSABLE_ENTITY)
+                .body(new HttpErrorInfo(HttpStatus.UNPROCESSABLE_ENTITY.value(), ex.getMessage()));
+    }
```
RichardDanon, line 61: Yeah that's better

RichardDanon (approved):
Everything looks good to go

PR #397, opened 2023-09-18
Cstm   add pet id/UUID   cpc 791

Nmazza1:
JIRA:
https://champlainsaintlambert.atlassian.net/browse/CPC-791?atlOrigin=eyJpIjoiZWU3MTFjNWQ0ZmYwNDAxYjhlODUxNWQyNTI4MDA3ODIiLCJwIjoiaiJ9
## Context: 
This commit adds a public identifier for the Pet Model and a private databases identifier.
## Changes
-Added public petId
-Changed Pet Testing Methods
## Dev notes (Optional)
The tests all pass so there should not be any issues when handling CRUD operations on Pets, if there is come tell me so I can send corresponding fixes.


DisgruntledMensch (approved):
looks good

customers-service-reactive/src/main/java/com/petclinic/customersservice/customersExceptions/http/HttpErrorInfo.java
```diff
@@ -25,20 +27,4 @@ public HttpErrorInfo() {
         this.message = null;
     }
 
-    public ZonedDateTime getTimestamp() {
```
mattjulienn, line 28: In all of our projects with Christine, we always have this usually so I'm not sure whether or not this is a 100% prerequisite but definitely something I'd keep for exception handling, but that's just me...

mattjulienn (changes requested):
read the comment, family

customers-service-reactive/src/main/java/com/petclinic/customersservice/customersExceptions/http/HttpErrorInfo.java
Nmazza1:
```
  public ZonedDateTime getTimestamp() {
        return this.timestamp;
    }
```
There is a method that pulls a timestamp

mattjulienn: looks good then, brother

mattjulienn (approved):
looks great, family

PR #411, opened 2023-09-21
bug(BILL-CPC-743): made date visible and change Date type to LocalDate

HennaCH:
JIRA: [link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-743)
## Context:
I needed to make the date visible when looking at the bills
## Changes

- I changed the name of the date variable from visitsDate to date, to make it match the name that was in BillDetails in the API-Gateway.

- I changed the date type to LocalDate()
- I changed the UI to only say Date 

- We can now see the date.

- The "Date\Time" subtitle is now changed to "Date".

Before:
Screenshot (885)

After:
Screenshot (886)


drallo22 (approved):
Good idea changing Date to LocalDate, also good idea setting the time zone to the system's default so that no matter where you are the time will be correct for you.

PR #415, opened 2023-09-21
feat/(VIST-CPC-762) User can now navigate to create new visit page

MichaelVacca:
JIRA: https://champlainsaintlambert.atlassian.net/jira/software/c/projects/CPC/boards/35?selectedIssue=CPC-762

Context:
I added a create visit button which when clicked will display a form for adding a new visit including a calendar

Changes
- Added a "Create Visit" Button
- Added CSS to the button and the H3 "Upcoming Visits" on the visit-list page
- Added routing to the "visitsNew" when the button is clicked
- I also temporarily changed the url for the create visit page since there is still work to be done for the old URL to make it work

Before the Change there was no "Create Visit" button and there was no way to access the create new visit page:
Screenshot 2023-09-21 010322

This is how the UI looks for the visit-list page
AfterChange

This is the create new visit page that the button navigates the user to
NewPage

### DIFF
--- a/api-gateway/src/main/resources/static/scripts/visit-list/visit-list.template.html
+++ b/api-gateway/src/main/resources/static/scripts/visit-list/visit-list.template.html
@@ -1,4 +1,43 @@
-<h3 style="margin-top: 18px">Upcoming Visits</h3>
+<style>
+
+    .container {
+        display: flex;
+        justify-content: space-between;
+        align-items: center;             /* This aligns items vertically */
+        width: 100%;
+        margin-bottom: 20px;
+    }
+
+    .original-button {
+        display: flex;
+        align-items: center;
+        justify-content: center;
+        line-height: 1;
+        text-decoration: none;
+        color: #ffffff;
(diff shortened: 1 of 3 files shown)


api-gateway/src/main/resources/static/scripts/visit-list/visit-list.template.html
```diff
@@ -1,4 +1,43 @@
...
+
+    .container {
+        display: flex;
+        justify-content: space-between;
+        align-items: center;             /* This aligns items vertically */
+        width: 100%;
+        margin-bottom: 20px;
+    }
+
+    .original-button {
```
swafit, line 11: none of this styling was demonstrated in the UI changes

swafit (changes requested):
Add a before UI changes image

swafit (approved):
[Nitpick]: This styling should be in the css class files.

TheyCallMeCris (approved):

PR #418, opened 2023-09-22
Feat/bill cpc 749 search filter by name

dahlica12:
JIRA: link to jira ticket
https://champlainsaintlambert.atlassian.net/browse/CPC-749 
## Context:
## The user is now able to retrieve bills from the bill history by the owner id, the vet id or both, which was previously not able to because the bill id had numbers that were part of the owner and vet id so the search was not specific enough. 
- The bill-history.template.html has been updated with two input fields for Filter by Owner Id and Filter by vet Id that bind the input entered to be synchronized with the respective id property. 
- The bill-history.js has been updated with two test cases for the expected value of owner id and vet id when a strict comparison is made with the filter. 
Before:
image
After:
image

## Dev notes (Optional)
- pull request link


drallo22 (approved):
Good way to filter by the id's will be very useful on the website! Code is simple but efficient.

HennaCH (approved):

HennaCH (commented):
The fact that we can now choose what to filter gives the filter its whole purpose. Good job !

PR #439, opened 2023-09-24
bug(VETS-CPC-792) Fix add vet

Sprytte:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-792
## Context:
This bug is fixed so that a user can add a vet through the add vet form.
## Changes

- Changed the value of the phone number in the vet-form.controller.js since it was trying to get the base phone number from a null value

## Dev notes (Optional)
- pull request link


DenisaNicoletaH (approved):
It looks good, the bug fix looks well done. The naming convention looks good as well. Overall, i think what you did is well done.

KarinaSofia (approved):
Good naming convention.

PR #451, opened 2023-09-25
Bug/bill cpc 746 change display id to names

ebond07:
JIRA: link to jira ticket
## Context:
The main page for bills (bill history) displayed a user Id and vet Id which meant nothing (they aren't the ones used as parameters in requests). So I switched them to instead display the owner's and vet's full names. Also had to change the type of these two attributes to match the type of their corresponding attribute in their respective microservice. This was done to facilitate later changes involving the two attributes.
## Changes
- Owner Name column added
- Owner Name column now displays owner's full name 
- Vet ID column added
- Vet Name column now displays vet's full name 
- CustomerId is now a string instead of an integer throughout the billling service and api-gateway
- VetId is like customerId now a string instead of an integer.
- Added a method to get the vet entity through matching the vetBillId attribute of the vet class and the vetId attribute of the bill class
- Added a method to get the customer's name through mapping the names with an id and then comparing that id to the customerId bill holds
Before
Screenshot 2023-09-24 201841
After
Screenshot 2023-09-25 193929

### DIFF
--- a/api-gateway/src/main/java/com/petclinic/bffapigateway/dtos/Bills/BillDetails.java
+++ b/api-gateway/src/main/java/com/petclinic/bffapigateway/dtos/Bills/BillDetails.java
@@ -1,22 +1,16 @@
 package com.petclinic.bffapigateway.dtos.Bills;
 
-import com.fasterxml.jackson.annotation.JsonIgnore;
 import lombok.Builder;
 import lombok.Data;
 
 import java.time.LocalDate;
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.List;
-
-import static java.util.stream.Collectors.toList;
 
 @Builder
 @Data
 public class BillDetails {
     private String billId;
     private LocalDate date;
-    private int customerId;
+    private String customerId;
     private String vetId;
     private String visitType;
     private double amount;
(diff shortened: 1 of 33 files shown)


api-gateway/src/main/java/com/petclinic/bffapigateway/dtos/Bills/BillDetails.java
```diff
@@ -1,22 +1,16 @@
...
-
-import static java.util.stream.Collectors.toList;
 
 @Builder
 @Data
 public class BillDetails {
     private String billId;
     private LocalDate date;
-    private int customerId;
+    private String customerId;
```
drallo22, line 13: Good job changing it to a string makes so much more sense.

api-gateway/src/main/resources/static/scripts/bill-history/bill-history.controller.js
```diff
@@ -6,6 +6,26 @@ angular.module('billHistory')
         let self = this;
         self.billHistory = []
 
+        self.owners = [
```
drallo22, line 9: If hardcoding was the only way to solve this problem, maybe we should consider using an aggregator

drallo22 (approved):
Looks good! As I said in the comments maybe we should consider using an aggregator, but looks good for what you were able to do considering the circumstances!

dahlica12 (approved):
Very useful feature to have, UI is more user-friendly and readable

HennaCH (approved):
The front end is more user friendly now, good job!

PR #454, opened 2023-09-25
feat(VIST-CPC-762): implemented create visit functionality

MichaelVacca:
JIRA: link to jira ticket: https://champlainsaintlambert.atlassian.net/jira/software/c/projects/CPC/boards/35?selectedIssue=CPC-762
## Context:
## Changes
- API gateway now uses string for petID and practionerId to work with the new changes in the back end
- Also removed a table in the new visit page since it does not display any data
- Post also uses LocalDateTIme

- Before Change 
BEforeNewTable
- After Change
Screenshot 2023-09-25 161916

## Dev notes (Optional)
- There is a known bug in which when a new visit is created, the request from the back end is sending the current date instead of the date chosen
- Also the response from the back-end returns null for the date
- These bugs don't effect the actual functionality of the visits page since we are still able to actually create a new visit and display it


Iantomasi (approved):
Good work getting the create functionality working Michael. I see what could be a potential fix for the front-end bug which could be the use of self.date in visits-templat.html instead of localdatetime. That's something we could try out next sprint.

SandwichKing978 (approved):
looks good

PR #471, opened 2023-09-28
feat(AUTH-CPC-889): auth implement get all users for the admin dashboard

DylanBrass:
JIRA: [link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-889?atlOrigin=eyJpIjoiYzkwODA1MTZiZGQ3NDViOWJiNjdmODgzMTA4YWM1MjgiLCJwIjoiaiJ9)
## Context:
This ticket was about implementing the get all users and adding UUID to the user. I took the time to test and achieve 90% coverage in the AUTH-service.
## Changes
-UUID implemented
-Get all implemented
-Admin dashboard is only visible to admin role
-90% coverage in AUTH-service
#### Before
image

#### After
image

image

### DIFF
--- a/api-gateway/src/main/java/com/petclinic/bffapigateway/dtos/Auth/UserDetails.java
+++ b/api-gateway/src/main/java/com/petclinic/bffapigateway/dtos/Auth/UserDetails.java
@@ -13,11 +13,8 @@ import java.util.Set;
 @Builder(toBuilder = true)
 public class UserDetails {
 
-    private long id;
-
+    private String userId;
     private String username;
-
-    private String password;
 
     private String email;
 
--- a/api-gateway/src/main/java/com/petclinic/bffapigateway/presentationlayer/BFFApiGatewayController.java
+++ b/api-gateway/src/main/java/com/petclinic/bffapigateway/presentationlayer/BFFApiGatewayController.java
@@ -344,10 +344,7 @@ public class BFFApiGatewayController {
 //    public Mono<UserDetails> getUserDetails(final @PathVariable long userId) {
 //        return authServiceClient.getUser(userId);
 //    }
-//    @GetMapping(value = "users")
-//    public Flux<UserDetails> getAll(@RequestHeader(AUTHORIZATION) String auth) {
-//        return authServiceClient.getUsers(auth);
-//    }
+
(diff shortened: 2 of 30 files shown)


auth-service/src/main/java/com/auth/authservice/datalayer/roles/Role.java
```diff
@@ -21,18 +21,10 @@
...
-    }
 
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     private long id;
     private String name;
 
-    @ManyToOne
-    @JoinColumn(name = "parent_id")
-    private Role parent;
```
RichardDanon, line 37: Great now that the structure of roles is completely flat we have more freedom in our usage of the role.

api-gateway/src/main/java/com/petclinic/bffapigateway/dtos/Auth/UserDetails.java
```diff
@@ -13,12 +13,9 @@
 @Builder(toBuilder = true)
 public class UserDetails {
 
-    private long id;
-
+    private String userId;
```
drallo22, line 16: Good change of making the userId a string instead of a long

api-gateway/src/main/java/com/petclinic/bffapigateway/presentationlayer/BFFApiGatewayController.java
```diff
@@ -482,13 +479,16 @@ public Mono<UserPasswordLessDTO> createUser(@RequestBody Register model) {
         return authServiceClient.createUser(model);
     }
 
+    @SecuredEndpoint(allowedRoles = {Roles.ADMIN})
+    @GetMapping(value = "users", produces = MediaType.TEXT_EVENT_STREAM_VALUE)
```
drallo22, line 483: Good use of text event stream as it is a flux

RichardDanon (approved):
Everything looks good. I would only say don't do the tests that concerns other stories let them handle it on their own.

drallo22 (approved):
All around good changes, get all users in the dashboard will be very useful and is a good add to the site!

PR #473, opened 2023-09-29
feat(AUTH-CPC-952):Create user specific endpoints

DylanBrass:
JIRA: [link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-952?atlOrigin=eyJpIjoiYWI3ZTcxY2FiMjY4NDYzMzk0NTU1YjI1MTRlNDAxY2IiLCJwIjoiaiJ9)
## Context:
This ticket was made after many requests from the vets and owners teams to make some endpoints only accessible by the concerned user. The changes add an annotation that makes it easy to secure and endpoint to be user specific. This means that the users id in the request must be the same stored in the JWE.
## Changes
-Added annotation
-Made User accounts for each owner.
-Made owner have static UUID
-Made a filter to check the @IsUserSpecific annotation.
-Linked the user and owner creation
-Changed the way edit and register works by changing the frontend

### Before : 
image
image
image

### After : 
image
image

After clicking edit account:
image

Clicking on another users profile gives this : 
image

## Dev notes (Optional)
-  The annotation has documentation, please read it.

### DIFF
--- a/api-gateway/src/main/java/com/petclinic/bffapigateway/utils/Security/Filters/IsUserFilter.java
+++ b/api-gateway/src/main/java/com/petclinic/bffapigateway/utils/Security/Filters/IsUserFilter.java
@@ -7,6 +7,7 @@ import com.petclinic.bffapigateway.exceptions.InvalidTokenException;
 import com.petclinic.bffapigateway.utils.Security.Annotations.IsUserSpecific;
 import com.petclinic.bffapigateway.utils.Security.Annotations.SecuredEndpoint;
 import com.petclinic.bffapigateway.utils.Security.Variables.Roles;
+import lombok.Generated;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.java.Log;
 import lombok.extern.slf4j.Slf4j;
@@ -30,6 +31,7 @@ import java.util.concurrent.ExecutionException;
 @Component
 @RequiredArgsConstructor
 @Order(3)
+@Generated
 public class IsUserFilter implements WebFilter {
 
     private final JwtTokenUtil jwtTokenUtil;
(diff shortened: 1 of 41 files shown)


api-gateway/src/main/java/com/petclinic/bffapigateway/utils/Security/Filters/IsUserFilter.java
```diff
@@ -0,0 +1,157 @@
...
+            throw new HandlerIsNullException("Handler is null, check if the endpoint is valid");
+        }
+
+        if (handler.getMethod().getAnnotation(IsUserSpecific.class) == null) {
+            return chain.filter(exchange);
+        }
+
+
+        if (Arrays.stream(handler.getMethod().getAnnotation(IsUserSpecific.class).bypassRoles()).anyMatch(role -> role == Roles.ALL)
+        || Arrays.stream(handler.getMethod().getAnnotation(IsUserSpecific.class).bypassRoles()).anyMatch(role -> role == Roles.ANONYMOUS)) {
```
Mila5847, line 69: Can you explain the filtering thing, please? This code:   if (exchange.getAttribute("whitelisted") != null && exchange.getAttribute("whitelisted") instanceof Boolean) {
            if((boolean) exchange.getAttribute("whitelisted")) {
                return chain.filter(exchange);
            }
        } got me particularly confused.

DylanBrass: The previous filters do a check if the request is anonymous, if it is, we allow everyone to accept it. To avoid repeating the check I added an attribute to the request (exchange) that would tell my future filters that the request has been white listed, it is also found in the roles filter.

Mila5847 (dismissed):

Mila5847 (dismissed):

DisgruntledMensch (dismissed):
Looks good to me, but ideally I would like another member of my team to review as well before merging just in case.

Nmazza1 (dismissed):
Not going to lie, this is a whole lotta Auth stuff I don't get, but now we can make a profile page for the currently signed in user!

Nmazza1 (approved):
Same Comment as last time.

DisgruntledMensch (approved):
all good

PR #489, opened 2023-10-01
Feat/vist cpc 788 change visit status to enum and cancel

CptDoritos:
JIRA: link to jira ticket
https://champlainsaintlambert.atlassian.net/browse/CPC-943
## Context:
Changed the Visit Status from a boolean to an enum and made a cancel button which changes the visit status to cancelled if possible so that visits would make more logical sense
## Changes
## Before
image
## After:
image


api-gateway/src/main/java/com/petclinic/bffapigateway/dtos/Visits/Status.java
```diff
@@ -0,0 +1,12 @@
+package com.petclinic.bffapigateway.dtos.Visits;
+
+public enum Status {
+
+    UPCOMING,
+    REQUESTED,
+    CONFIRMED,
+    CANCELLED,
+    IN_PROGRESS,
```
Iantomasi, line 9: Good stuff Fred, I. think we should maybe discuss removing one or two of the listed status option in here together for the next pull request

Iantomasi (approved):
This certainly passes the vibe check

AlexeiTimbro (approved):
I approve of these changes regarding the handling of the status of visits.

PR #490, opened 2023-10-01
feat(CSTM-CPC-857): add pagination functionality to current owners page

Youssefchahboune:
JIRA: https://champlainsaintlambert.atlassian.net/jira/software/c/projects/CPC/boards/14?selectedIssue=CPC-857&sprints=305
## Context:
This ticket is for implementing pagination to fetch a specified number of owners per page.
## Changes

- Added pagination endpoint and getTotalNumberOfOwners endpoint.
- Added tests for each of the added endpoints.
- Modified the owner-list controller to make the appropriate calls to the backend to update the URL and return the corresponding response and data.
- Implemented a UI component at the bottom of the table to allow switching between pages.

Before : 
image
After : 
image
image
image


Nmazza1 (approved):
I couldn't find the value in your changes, but if the default value is 5 per page, I personally think more on a page is fine, say 10 at most. (In a real environment with more data this would apply more)

api-gateway/src/main/java/com/petclinic/bffapigateway/presentationlayer/BFFApiGatewayController.java
```diff
@@ -437,6 +438,25 @@ public Flux<OwnerResponseDTO> getAllOwners() {
                                 .map(addVisitsToOwner(n))
                 );*/
     }
+    @GetMapping(value = "/owners-pagination")
+    public Flux<OwnerResponseDTO> getOwnersByPagination(@RequestParam Optional<Integer> page, @RequestParam Optional<Integer> size) {
+
+        if(page.isEmpty() || page == null){
+            page = Optional.of(0);
```
DylanBrass, line 445: Why is it 0, is it like an array where it starts at 0 ?

DylanBrass (approved):

api-gateway/src/main/java/com/petclinic/bffapigateway/presentationlayer/BFFApiGatewayController.java
Youssefchahboune: The pages start at 0, so if no page numbers have been specified, it returns page 0. These pages can be thought of as sections, starting from section zero, with each section containing a certain number of owners. if you look at the back end implementation you will see how it works.

PR #495, opened 2023-10-01
feat(CPC-896): Fix update inventory Form bug

SandwichKing978:
JIRA: link to jira ticket
https://champlainsaintlambert.atlassian.net/browse/CPC-896
## Context:

- Created a new form called inventory-update-form -form, 
- Added functionality for a user to update a specific inventory
- the form gets the Id of a specific inventory and displays the values in the form

## Changes

- I separated the update inventory form logic , from the inventory form by creating another folder called inventory-update
- My changes affect inventory management 

Before
image
After
image
image

Before, when a user clicked on the edit form they would get a form that would say add new inventory and the form wouldn't be prefilled with values.
Now, a user is directed to a form that says update form and the form is already filled with prefilled values from that inventory.


Matthewtan9 (dismissed):
Looks good to me. Great job on fixing the bug!

AlexeiTimbro (dismissed):
Everything looks good regarding the fixing of the update/edit button.

PR #503, opened 2023-10-02
feat(VIST-CPC-965): Implement Exception Handling For Status

TheyCallMeCris:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-965
## Context:
Exception Handling was done for every attribute of the request that is made when posting. However, we have changed Status from a boolean to an enum class. So changes had to be made concerning that.
## Changes
Added another else if in my validateRequestDTO method inside our service in order to check if the Status is REQUESTED.


MichaelVacca (approved):

CptDoritos (approved):
This passes the Vibe Check

PR #510, opened 2023-10-03
feat(AUTH-CPC-963): auth remove secret key to be only in auth service

DylanBrass:
JIRA: [link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-963?atlOrigin=eyJpIjoiNzg5ZTg4NzVhZTZiNGEzYjgzYzMwYTQ3NmYzY2U2ZDEiLCJwIjoiaiJ9)
## Context:
Removed stressed caused by checking roles in api-gateway and removed the stored secret in the gateway.
## Changes
-Removed secret in api-gateway
-Made new DTO to store all useful information in filters.


nic5694 (approved):
Tests are complete and well done

api-gateway/src/main/java/com/petclinic/bffapigateway/dtos/Auth/TokenResponseDTO.java
```diff
@@ -0,0 +1,21 @@
...
+@NoArgsConstructor
+@Builder(toBuilder = true)
+public class TokenResponseDTO {
+
+        private String token;
+        private String userId;
+        private String email;
+        private List<String> roles;
+
+}
```
RichardDanon, line 21: Why did you create a token DTO exactly?

RichardDanon (approved):
Everything in this code looks good!

api-gateway/src/main/java/com/petclinic/bffapigateway/dtos/Auth/TokenResponseDTO.java
DylanBrass: To reduce the numbers of call since like that we make sure we have everything in one request.

PR #511, opened 2023-10-03
feat(VIST-CPC-947): implement delete all canceled visits functionality

Iantomasi:
JIRA: link to jira ticket
https://champlainsaintlambert.atlassian.net/browse/CPC-947
## Context:
This ticket pertains to implementing a new functionality in the frontend that allows for the user to delete all cancelled visits from the database. This functionality compliments our teams newly incorporated Cancel Visit functionality.
## Changes
- deleteAllCancelledVisits method logic implemented in the backend
- successful and exception test cases implemented in the frontend and backend test classes
- UI styling touchups

Before changes:
image

After changes:
AfterChanges

AfterBtnPress1

Results

## Dev notes
- Our team has decided that we will be making some minor changes to the separation logic regarding the Upcoming and Previous visits. Will be introduced in upcoming pull requests.

## Linked pull requests 

[- pull request link](https://github.com/cgerard321/champlain_petclinic/pull/489)

### DIFF
--- a/api-gateway/src/main/resources/static/scripts/visit-list/visit-list.controller.js
+++ b/api-gateway/src/main/resources/static/scripts/visit-list/visit-list.controller.js
@@ -22,6 +22,14 @@ angular.module('visitList')
             }
         }
 
+        function delayedReload() {
+            var loadingIndicator = document.getElementById('loadingIndicator');
+            loadingIndicator.style.display = 'block';
+            setTimeout(function() {
+                location.reload();
+            }, 1000); //delay by 1 second
+        }
+
         $scope.cancelVisit = function (visitId, status){
             console.log("Called Function")
             console.log(status)
@@ -30,8 +38,6 @@ angular.module('visitList')
                 status = "CANCELLED"
             }else if(status === "CONFIRMED"){
                 status = "CONFIRMED"
-            }else if(status === "IN_PROGRESS"){
-                status = "IN_PROGRESS"
             }else if(status === "COMPLETED"){
                 status = "COMPLETED"
             }else{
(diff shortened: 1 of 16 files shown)


api-gateway/src/main/resources/static/scripts/visit-list/visit-list.controller.js
```diff
@@ -78,15 +85,28 @@ angular.module('visitList')
...
+                    delayedReload();
+                }
+
+                function errorCallback(error) {
+                    alert(data.errors);
+                    console.log(error, 'Could not receive data');
+                }
             }
         };
+
```
MichaelVacca, line 109: Can you explain this?

CptDoritos: > Can you explain this?

I know this isn't my pr, but this is the front-end method used by the button

Success callback and Error callback are the responses that are given if te function is successful or fails

CptDoritos (approved):
This definitely passes the vibe check

api-gateway/src/main/resources/static/scripts/visit-list/visit-list.controller.js
Iantomasi: Of course bellume! So this is defining the deleteAllCancelledVisits method in the front end that we then link our button to in visit-list.template.html.  It displays a confirmation pop-up after the button press and waits for the user to confirm before actually making a request to the "visits/cancelled" uri that we defined in the backend. If this is successful it calls the successCallback method that displays a message as well as calls our delayedReload method that will refresh our data. If there would be an issue, it would display 'Could not receive  data' instead.

MichaelVacca (approved):

PR #526, opened 2023-10-06
Feat/INVT-CPC-889 search letter or number result

Matthewtan9:
JIRA: link to jira ticket:
https://champlainsaintlambert.atlassian.net/browse/CPC-899
## Context:
Added the ability to search a product by a letter or a number
## Changes
-After further review with my teammate I have removed all the unnecessary code unused.
-Modified the back-end for Inventory and product get all to regex which enables the inventory to appear depending on the letter and number inputed.
-Added a button that will facilitate the search better which is the Enter key button rather than just clicking the search

Before (By the entire name needed to be written):
image

Before searching by letter or number for inventory:
image
After putting a letter:
image
After inserting a number: 
image
For description: 
image

Before Product:
image

After inserting the product letter or numeric digit for the name:
image
image

## Dev notes (Optional)
-Did not implement this feature for the product price and product quantity because regex does not support double or Int. If I were to replace all double or int to string it would affect and create a bunch of merge conflicts.I would also need to change all the database double and int type to string too.
- pull request link

### DIFF
--- a/api-gateway/src/main/resources/static/scripts/inventory-list/inventory-list.controller.js
+++ b/api-gateway/src/main/resources/static/scripts/inventory-list/inventory-list.controller.js
@@ -13,6 +13,13 @@ angular.module('inventoryList')
 
         getInventoryList()
 
+
+        $http.get('api/gateway/inventory').then(function (resp) {
+            self.inventoryList = resp.data;
+            console.log("Resp data: " + resp.data)
+            console.log("inventory list: " + self.inventoryList)
+        });
+
         $scope.searchInventory = function (inventoryName, inventoryType, inventoryDescription){
             getInventoryList(inventoryName, inventoryType, inventoryDescription)
         }
@@ -27,10 +34,11 @@ angular.module('inventoryList')
 
             if (inventoryName != null && inventoryName !== '') {
                 name = inventoryName
+
                 queryString += "inventoryName=" + inventoryName;
             }
 
-            if (inventoryType != null && inventoryType !== '') {
+            if (inventoryType) {
                 if (queryString !== '') {
                     queryString += "&";
                 }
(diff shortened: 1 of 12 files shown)


api-gateway/src/main/resources/static/scripts/inventory-list/inventory-list.controller.js
```diff
@@ -13,6 +13,13 @@ angular.module('inventoryList')
 
         getInventoryList()
 
+
+                $http.get('api/gateway/inventory').then(function (resp) {
+                    self.inventoryList = resp.data;
+                    console.log("Resp data: " + resp.data)
+                    console.log("inventory list: " + self.inventoryList)
+                });
+
```
emilegirardGit, line 22: I think this might break the pagination because you need to add the query params for the page and size of the list. Can you tell me if the pagination still work

Matthewtan9: image
image
After running your code. The pagination that you made still works fine.

emilegirardGit: If it works great, but I don't think you need to keep it. The getInventoryList() is called without any field so it is doing the get all inventories.

emilegirardGit (dismissed):

api-gateway/src/main/resources/static/scripts/inventory-list/inventory-list.controller.js
Matthewtan9: I will take it into consideration and talk with the team members to decide whether or not we should remove it.
Thank you.

AlexeiTimbro (approved):
Happy to hear that you removed the necessary code. Everything looks good code wise keep it up.

Eli293 (approved):

PR #529, opened 2023-10-06
feat(BILL-CPC-922): Add "due date" for bills

MinseokBUZZ:
JIRA: link to jira ticket
## Context:

Initially, there was no due date which provides clarity about when payment is expected. 
So as a developer, thought it would be much better to specify the deadline which notifies client as soon as possible.

## Changes

- Added "due date" using LocalDate in api-gateway as well as billing-service
- Added "due date" variable on tests for both api-gateway & billing-service

**Before**
image
**After**
image
image
image

## Dev notes (Optional)
- pull request link


api-gateway/src/main/resources/static/scripts/bill-details/bill-details.template.html
```diff
@@ -38,6 +38,9 @@ <h5>Address: 505 Lincoln Str. Queens, New York</h5>
             <h5>Bill Status: <span ng-style="{'color': $ctrl.bills.billStatus === 'OVERDUE' ? 'red' : 'black'}">{{$ctrl.bills.billStatus}}</span></h5>
             <hr>
             <br>
+            <h5>Due Date: <strong>{{$ctrl.bills.dueDate}}</strong></h5>
```
drallo22, line 41: I like the use of <strong> as it makes the due date "pop" more which is good because the due date is important!

api-gateway/src/test/java/com/petclinic/bffapigateway/domainclientlayer/BillServiceClientIntegrationTest.java
drallo22, line 1: Why are the due dates being set to null?

billing-service/src/main/java/com/petclinic/billing/datalayer/DataSetupService.java
```diff
@@ -22,11 +22,11 @@ public DataSetupService(BillService billService) {
     public void run(String... args) throws Exception {
 
 
-        BillRequestDTO b1 = new BillRequestDTO( "1", "general", "1", LocalDate.of(2021,9,19),59.99,BillStatus.PAID);
-        BillRequestDTO b3 = new BillRequestDTO( "3", "operation", "1", LocalDate.of(2021,9,21), 199.99,BillStatus.PAID);
-        BillRequestDTO b4 = new BillRequestDTO( "4", "injury", "1",  LocalDate.of(2021,9,22), 199.99,BillStatus.UNPAID);
-        BillRequestDTO b2 = new BillRequestDTO( "2", "operation", "2", LocalDate.of(2021,9,20), 199.99,BillStatus.OVERDUE);
-        BillRequestDTO b5 = new BillRequestDTO( "5", "chronic", "3",  LocalDate.of(2021,9,23), 199.99,BillStatus.UNPAID);
+        BillRequestDTO b1 = new BillRequestDTO( "1", "general", "1", LocalDate.of(2023,9,19),59.99,BillStatus.PAID, LocalDate.of(2023, 10,3));
```
drallo22, line 25: Good idea of updating the dates of the bills to more recently so that the due dates make more sense!

billing-service/src/main/java/com/petclinic/billing/datalayer/DataSetupService.java
```diff
@@ -22,11 +22,11 @@ public DataSetupService(BillService billService) {
...
+        BillRequestDTO b3 = new BillRequestDTO( "3", "operation", "1", LocalDate.of(2023,9,21), 199.99,BillStatus.PAID,LocalDate.of(2023, 10,5));
+        BillRequestDTO b4 = new BillRequestDTO( "4", "injury", "1",  LocalDate.of(2023,9,30), 199.99,BillStatus.UNPAID,LocalDate.of(2023, 10,10));
```
drallo22, line 27: I think this might be a small error here as the due date and bill date are not 14 days apart

drallo22 (dismissed):
Efficient and useful code! This ties in nicely with the bill status and will be very useful for employees.

api-gateway/src/test/java/com/petclinic/bffapigateway/domainclientlayer/BillServiceClientIntegrationTest.java
MinseokBUZZ: When I put local date on due date, it gives me the error that says `**java.time.LocalDate` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling (through reference chain: com.petclinic.bffapigateway.dtos.Bills.BillDetails["dueDate"])**. So i checked the build.gradle to put the dependency for localdate, but it's already there. 
image
So, that's why the one of our team members put date as "NULL" as well. So I decided to leave the due date as "NULL" so that i don't break the code for now and be able to fix both date and due date variables later on.

billing-service/src/main/java/com/petclinic/billing/datalayer/DataSetupService.java
MinseokBUZZ: Thank you for letting me know even though it's a small error. it's fixed and they are now 14 days apart like others.

HennaCH (approved):
I think that it is a good idea of adding a due date, specially making it a different color depending on whether it is overdue or not. Good job !

drallo22 (approved):
Fixed all problems I mentioned, great code!

PR #505, opened 2023-10-03
feat(INVT CPC-894): Ability to add and search inventory type

emilegirardGit:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-894
## Context:
When you want to search for an inventory type if the inventory type is not found it create the new type
## Changes
- Modify inventory type from an enum to a repository
- Added post method
- Change front end dropdown
- Added search bar
## Before and After UI (Required for UI-impacting PRs)
InventroyTypeAfter1
InventoryTypeAfter2
InventoryTypeAfter3
InventoryTypeAfter4

PR #530, opened 2023-10-07
feat(INVT-CPC-895): Get all custom inventory types

Eli293:
JIRA: link to jira ticket
https://champlainsaintlambert.atlassian.net/browse/CPC-895
## Context:
This ticket is about adding the rendering of all created inventory types after creating them.
## Changes

- This affects the inventory
- This affects the addition and editing of inventories
- Added get all inventory type request
- Displays all the default + custom inventory types in the dropdown bar
- Search queries now support the custom inventory types as well + displayed when clicked on
- All testing required is completed

Before
image
After
image
Add Form
image
Update Form
image

https://github.com/cgerard321/champlain_petclinic/pull/505

### DIFF
--- a/api-gateway/src/main/resources/static/scripts/Inventory-form/inventory-form.controller.js
+++ b/api-gateway/src/main/resources/static/scripts/Inventory-form/inventory-form.controller.js
@@ -5,7 +5,12 @@ angular.module('inventoryForm')
         var self = this;
         console.log("State params: " + $stateParams)
         $scope.inventoryTypeFormSearch = "";
-        $scope.inventoryTypeOptions = ["New Type", "Internal", "Sales"] //get all form the inventory type repository but dont remove the New Type
+        $scope.inventoryTypeOptions = ["New Type"] //get all form the inventory type repository but dont remove the New Type
+        $http.get("api/gateway/inventory/type").then(function (resp) {
+            //Includes all types inside the array
+            resp.data.forEach(function (type) {
+                $scope.inventoryTypeOptions.push(type.type);
+            })});
         $scope.selectedOption = $scope.inventoryTypeOptions[0]
 
         self.submitInventoryForm = function () {
--- a/inventory-service/src/main/java/com/petclinic/inventoryservice/presentationlayer/InventoryController.java
+++ b/inventory-service/src/main/java/com/petclinic/inventoryservice/presentationlayer/InventoryController.java
@@ -128,5 +128,10 @@ public Flux<InventoryResponseDTO> searchInventories(
                 .defaultIfEmpty(ResponseEntity.notFound().build());
     }
 
+    @GetMapping("/type")
+    public Flux<InventoryTypeResponseDTO> getAllInventoryTypes(){
+    return productInventoryService.getAllInventoryTypes();
+    }
+
 }
 
--- a/api-gateway/src/main/java/com/petclinic/bffapigateway/domainclientlayer/InventoryServiceClient.java
+++ b/api-gateway/src/main/java/com/petclinic/bffapigateway/domainclientlayer/InventoryServiceClient.java
@@ -197,6 +197,14 @@ public class InventoryServiceClient {
                 .retrieve().bodyToMono(InventoryTypeResponseDTO.class);
     }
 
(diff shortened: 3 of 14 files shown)


api-gateway/src/main/resources/static/scripts/Inventory-form/inventory-form.controller.js
```diff
@@ -5,7 +5,12 @@ angular.module('inventoryForm')
         var self = this;
         console.log("State params: " + $stateParams)
         $scope.inventoryTypeFormSearch = "";
-        $scope.inventoryTypeOptions = ["New Type", "Internal", "Sales"] //get all form the inventory type repository but dont remove the New Type
+        $scope.inventoryTypeOptions = ["New Type"] //get all form the inventory type repository but dont remove the New Type
```
Matthewtan9, line 8: Mind explaining this code and why you removed Internal and Sales?

nic5694 (commented):
When you click new type does it open a form or how does it work?

api-gateway/src/main/resources/static/scripts/Inventory-form/inventory-form.controller.js
Eli293: No problem, since I am getting all inventory types, there is no need to hard code Internal and Sales inside the array, since it is all supposed to be fetched with the get all inv types request. We keep "New Type" since it is tied to the add inventory type request functionality.

Matthewtan9: I appreciate the response. Good job on the task.

Matthewtan9 (approved):
After further further review this code looks great.

inventory-service/src/main/java/com/petclinic/inventoryservice/presentationlayer/InventoryController.java
```diff
@@ -128,5 +128,10 @@ public Mono<ResponseEntity<InventoryTypeResponseDTO>> addInventoryType(@RequestB
                 .defaultIfEmpty(ResponseEntity.notFound().build());
     }
 
+    @GetMapping("/type")
```
nic5694, line 131: The GetMapping isn't off? Shouldn't it be /inventory/type?

inventory-service/src/main/java/com/petclinic/inventoryservice/presentationlayer/InventoryController.java
```diff
@@ -128,5 +128,10 @@ public Mono<ResponseEntity<InventoryTypeResponseDTO>> addInventoryType(@RequestB
...
+    public Flux<InventoryTypeResponseDTO> getAllInventoryTypes(){
```
nic5694, line 132: Shouldn't you be returning a ResponseEntity, with the 200 OK?

nic5694 (commented):
Please take a look at the comments i made, it was mostly to keep the API gateway as reactive as possible.

Eli293:
> When you click new type does it open a form or how does it work?

When "New Type" is selected, the textbox will be empty to fill out a custom inventory type. If nothing is entered, an alert will pop up. Once the new inventory type is created, the new type will now be displayed in the list of available types which you can perform adding, editing, and search operations with.

nic5694 (approved):

inventory-service/src/main/java/com/petclinic/inventoryservice/presentationlayer/InventoryController.java
Eli293: Nice observation, but the /inventory should already be included in the @RequestMapping at the beggining, adding /inventory/type would simply be redundant no?

inventory-service/src/main/java/com/petclinic/inventoryservice/presentationlayer/InventoryController.java
Eli293: When you return a Flux, the framework should take care of serializing the data into the appropriate format and setting the appropriate HTTP response headers, like the Content-Type. We should not need to create a ResponseEntity manually while using Flux unless I am clueless.

PR #532, opened 2023-10-08
Feat/invt cpc 981 access product list through inventory row

AlexeiTimbro:
JIRA: link to jira ticket
https://champlainsaintlambert.atlassian.net/browse/CPC-989       
                   
What is the ticket about and why are we making this change?

In this pull request, I made tweaks to the front end to enrich the user experience. Instead of clicking on the name of the inventory to view the product inside of it now you can simply click on the row itself with the hover on the mouse feature that I added as well. We are making this change because it makes the flow of the inventory more user-friendly and easier to navigate through the inventory.

What are the various changes and what other modules do those changes affect?

-Clicking on the row of an inventory links you to the product list inside of that inventory
-CSS: added hover on the mouse that appears on the inventory rows
-Made the edit and delete buttons inside the rows work properly with the new feature I added

Before: 
image

After: Mouse Hover feature
image

After: clicking anywhere on the selected row brings you to the Product List
image

## Dev notes (Optional)
- We go Gym


Eli293 (approved):

Iantomasi (approved):
Like the greying out of the row when your cursor hovers it, think it looks sleek and modern.

PR #533, opened 2023-10-09
Feat/invt cpc 979 add sale price for clients in product list

AlexeiTimbro:
JIRA: link to jira ticket
https://champlainsaintlambert.atlassian.net/browse/CPC-979

What is the ticket about and why are we making this change?

This ticket is about adding an important field to our product list called the sale price. The sale price is important because previously we only had the normal price for a product that the pet clinic bought it for and we were missing the price we will be selling it at. We are making those changes because it's crucial information about a product.

What are the various changes and what other modules do those changes affect?

- The front end has been changed to accommodate the new field in the product list.
- A couple of things about the backend have been changed: 
1. RequestDTO 
2. ResponseDTO
3. Product Entity
4. ServiceImpl methods

- All the Testing that was using the DTOs, the Product entity and the serviceImpl has been changed to accommodate the new changes.
- API Gateway testing fully accommodated with the new changes.

Before: 
image

After:
image

### DIFF
--- a/api-gateway/src/main/java/com/petclinic/bffapigateway/dtos/Inventory/ProductRequestDTO.java
+++ b/api-gateway/src/main/java/com/petclinic/bffapigateway/dtos/Inventory/ProductRequestDTO.java
@@ -14,4 +14,6 @@ public class ProductRequestDTO {
     private String productDescription;
     private Double productPrice;
     private Integer productQuantity;
+    private Double productSalePrice;
+
 }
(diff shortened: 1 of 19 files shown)


Matthewtan9 (approved):
Code looks good overall. Good job.

Iantomasi (approved):
Looks good, simple object value addition and logic has been updated acoordingly. UI shows the value.

Matthewtan9 (approved):

Iantomasi (approved):

PR #536, opened 2023-10-10
feat(VIST-CPC-913): sorting visits into status tables

Iantomasi:
JIRA: link to jira ticket
https://champlainsaintlambert.atlassian.net/browse/CPC-913

## Context:
This ticket is regarding making updates to the Visit-List UI. Our team thought it would be an interesting idea to separate visits into more than just a single table and, to compliment our recent implementations of status-button functionalities, we thought it would be best for these tables to be based on visit status. We now have a total of four tables: Upcoming Visits, Confirmed Visits, Cancelled Visits and Completed Visits. 

## Changes
- Tweaks to the controller.js file were made to push visitData to the appropriate status table
- Three new status tables were added to the html template (aside from our original upcoming visits table that was previously implemented)
- Styling modifications were made to make the UI maintain consistency and be more visually appealing

Before changes: 
273456000-c9c47dd1-d8da-44cc-b0ca-7e873fdfa0ad

After Changes:
After_Changes

After all visits are removed from a particular table:
After_Removing_All_Visits_From_A_Table

## Dev notes (Optional)
- A table that has no visits in it will be hidden until a new one is eventually added to it. This is possible through the use of ' ng-show="$ctrl.table.length > 0 '

### DIFF
--- a/api-gateway/src/main/resources/static/scripts/visit-list/visit-list.controller.js
+++ b/api-gateway/src/main/resources/static/scripts/visit-list/visit-list.controller.js
@@ -4,7 +4,10 @@ angular.module('visitList')
         let self = this;
         // Lists holding visits for the tables to display
         self.upcomingVisits = []
-        self.previousVisits = []
+        self.confirmedVisits = []
+        self.cancelledVisits = []
+        self.completedVisits = []
+
         let url
         //sorted by order or most to least permissions
         if ($window.localStorage.getItem("roles").includes("ADMIN")){
@@ -19,9 +22,25 @@ angular.module('visitList')
         eventSource.addEventListener('message', function (event){
             $scope.$apply(function(){
                 console.log(event.data)
-                self.upcomingVisits.push(JSON.parse(event.data))
-            })
-        })
+                const visitData = JSON.parse(event.data);
+                switch (visitData.status) {
+                    case "UPCOMING":
+                        self.upcomingVisits.push(visitData);
+                        break;
+                    case "CONFIRMED":
+                        self.confirmedVisits.push(visitData);
+                        break;
+                    case "CANCELLED":
+                        self.cancelledVisits.push(visitData);
+                        break;
+                    case "COMPLETED":
+                        self.completedVisits.push(visitData);
+                        break;
+                    default:
+                        break;
+                }
+            });
+        });
         eventSource.onerror = (error) =>{
             if(eventSource.readyState === 0){
                 eventSource.close()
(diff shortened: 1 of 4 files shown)


swafit:
Could the filters be added to all the tables?

swafit:
To avoid duplicating the tables multiple times, would it be better to have 1 visits table and add a dropdown to select the status instead of a filter to sort for only 1 type of visit?

api-gateway/src/main/resources/static/scripts/visit-list/visit-list.template.html
swafit, line 1: [Nitpick]: the styling should be transferred to a CSS StyleSheet

swafit (approved):

api-gateway/src/main/resources/static/scripts/visit-list/visit-list.controller.js
```diff
@@ -19,9 +22,25 @@ angular.module('visitList')
         eventSource.addEventListener('message', function (event){
             $scope.$apply(function(){
                 console.log(event.data)
-                self.upcomingVisits.push(JSON.parse(event.data))
-            })
-        })
+                const visitData = JSON.parse(event.data);
+                switch (visitData.status) {
+                    case "UPCOMING":
+                        self.upcomingVisits.push(visitData);
```
swafit, line 28: Sorting was not modified for the new lists. If you select any sorting button on any table, it will only modify the sorting for upcoming visits.

api-gateway/src/main/resources/static/scripts/visit-list/visit-list.template.html
Iantomasi: You are correct Will-I-Am

api-gateway/src/main/resources/static/scripts/visit-list/visit-list.controller.js
Iantomasi: I see, I'll speak to you in person to make sure my understanding of the problem is correct

CptDoritos (approved):
This passes the Vibe Check

Iantomasi:
> Could the filters be added to all the tables?

I believe we discussed this in the previous scrum but I could be mistaken. For the filters, if one is in use then it automatically filters all the visits we are searching for in all tables it is present in (if for example we would search by a date, then all visits with that date will be highlighted in all tables in the UI). I thought it would maybe be unnecessary to have the same search filters on each table but we'll discuss it and could add it again in a future pr if.

PR #560, opened 2023-10-20
bug(INVT-CPC-1009):Fixed the list issue in the docker-compose

nic5694:
Took 3 minutes

JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-1009
## Context:
Fixed the bug with the lists when you try and bring the service up or build.
## Changes
- Removed the commented out ports and removed the ports section.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -42,8 +42,6 @@ services:
 
     customers-service-reactive:
         build: customers-service-reactive
-        ports:
-#            - "8090:8080"
         environment:
             - SPRING_PROFILES_ACTIVE=docker
         depends_on:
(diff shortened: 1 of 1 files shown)


Nmazza1 (approved):
Good stuff, glad we could get Main back online! 👍

docker-compose.yml
```diff
@@ -42,8 +42,6 @@ services:
 
     customers-service-reactive:
         build: customers-service-reactive
-        ports:
-#            - "8090:8080"
```
RichardDanon, line 46: Why did you do that anyways?

RichardDanon (approved):
Yes, thank you for fixing the issue.

docker-compose.yml
Nmazza1: 👍

nic5694: because @Nmazza1  was complaining about the ports so i wanted to keep it commented out so it would be easier for him. Clearly it was a mistake to try and make his life easier....

PR #561, opened 2023-10-20
Feat(BILL CPC-919)Delete all bills in bill history

dahlica12:
JIRA: link to jira ticket
https://champlainsaintlambert.atlassian.net/browse/CPC-919
## Context:
Have a delete all bills button and functionality for the option to delete all bills.
## Changes
- Added delete all bills button
- Created methods for delete all bills functionailty
-  Created tests for delete all bills methods

Before:
image
After:
image

## Dev notes (Optional)
- pull request link


HennaCH (dismissed):
Good implementation, good job !

api-gateway/src/main/resources/static/scripts/bill-history/bill-history.controller.js
```diff
@@ -133,6 +133,31 @@ angular.module('billHistory')
             }
         };
 
+
+        $scope.deleteAllBills = function () {
+            let varIsConf = confirm('Are you sure you want to delete all the bills in the bill history');
```
drallo22, line 138: Looks good! Another way you could have done this is by making it so that if the user does not confirm it, it would simply return out of the function but if the user does confirm it, all the code in the function is executed. Just a small difference but both work well!

drallo22 (approved):
Looks clean and efficient, good work!

api-gateway/src/main/resources/static/scripts/bill-history/bill-history.controller.js
dahlica12: Alright, I will try that out!

dahlica12: @drallo22 I took your suggestion and have made the change so in the function it will return out of the function if the user does not confirm the delete operation

HennaCH (approved):
Good way to let the user know what is happening with the data

drallo22 (approved):

api-gateway/src/main/resources/static/scripts/bill-history/bill-history.controller.js
drallo22: Nice!

PR #567, opened 2023-10-21
Feat (INVT-CPC 991):Ability to make the products row hoverable

SandwichKing978:
JIRA: link to jira ticket
https://champlainsaintlambert.atlassian.net/browse/CPC-991
## Context:

- I added a hover element to the inventory products row
- when u hover over the row you can see clearly which row you selected with the hover elements 

## Changes

- I removed the product details folder because it's a unused directory the code is never being used. 

I added a hover element to the inventory products row
- My changes affect products

Before 
image
After 
image


api-gateway/src/main/resources/static/scripts/product-details/product-details.component.js
emilegirardGit, line 1: Is it normal that the all the products details files are deleted

SandwichKing978: Yes because its an empty folder with code never being used, i created another folder where it gets the product details its called product details info.

Matthewtan9 (approved):
looks fine

Nmazza1 (approved):
I would recommend specifying in your PR why you deleted the file as it jarring opening the review to find a lot of deleted lines.

emilegirardGit (approved):

PR #571, opened 2023-10-22
Feat/cstm cpc 1013 add pet weight

1-emil:
JIRA: [link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-1013?atlOrigin=eyJpIjoiYjNhOTkwYmQzOTcyNDE4N2I4MmZlMzQzNzMwMjU1YTciLCJwIjoiaiJ9)
## Context:
Adding pet weight to front-end and back-end so the customer can edit the see their pet's weight and edit it
## Changes
- Add weight string
- Add weight to confirmation when submitting edit pet form
- add weight input text box to edit and add pet form
- add weight text box to pet details to see weight
Screenshot 2023-10-22 at 1 51 44 AM

I don't have a before image but its basically the form and confirmation message without the weight.
## Dev notes (Optional)
- I also fixed a small bug where isActive would be blank in the pet details after editing a pet.


DisgruntledMensch (approved):
Looks good, make sure all the testing is still working since you've added a new var.

Nmazza1 (commented):
Why are you adding a lot of code back to Owner Details after it was moved?

1-emil:
> Why are you adding a lot of code back to Owner Details after it was moved?

I did not add this code I think it added when i merged main with my branch

Nmazza1 (approved):
In that case I suggest deleting the code if possible to keep the files organized, otherwise all good 👍

PR #572, opened 2023-10-22
feat/BILL-CPC-933_Implement_Button_To_Calculate_Bill_With_Tax

HennaCH:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-933
## Context: 
We need to see the amount to pay including taxes
## Changes

- Added taxedAmount field in BillResponseDTO and Bill
- Added tax calculation in BillServiceImpl and BillServiceClient
- I modified the BillResponseDTP in the tests so that they include the taxedAmount
- Made the button to display the taxed amount 

Before
Screenshot (1130)

After
Screenshot (1144)
___________________________________________________________
Screenshot (1145)

## Dev notes (Optional)
- I am not sure if my changes in the test sections are good


api-gateway/src/main/resources/static/scripts/bill-details/bill-details.template.html
```diff
@@ -42,6 +42,8 @@ <h5>Due Date: <strong>{{$ctrl.bills.dueDate}}</strong></h5>
             <hr>
             <br>
             <h5>Final Amount Due: {{$ctrl.bills.amount}}</h5>
+            <h5 ng-if="show">Final Amount Including Taxes: {{$ctrl.bills.taxedAmount}}</h5>
+            <button class="newBtn btn btn-primary" ng-click="show = !show"> {{show? "Hide amount including taxes" : "Show amount including taxes"}}</button>
```
drallo22, line 46: You also could have done this by crating a style class that is hidden, and then when the button is clicked the style class is then shown! Just a small change, either way works well!

drallo22 (approved):

api-gateway/src/main/resources/static/scripts/bill-details/bill-details.template.html
HennaCH: Ok, I will try to do your suggestion !

billing-service/src/main/java/com/petclinic/billing/businesslayer/BillService.java
```diff
@@ -12,6 +12,8 @@
 public interface BillService {
     Mono<BillResponseDTO> getBillByBillId(String billId);
 
+    //Mono<BillResponseDTO> getTaxedAmountByBillIdAndBillAmount(String billId, Double amount);
+
```
ebond07, line 16: Is this left here so that eventually it will be implemented in the service impl? If not I don't think it would be a bad idea to get rid of commented out lines that we aren't using.

ebond07 (dismissed):
Good job adding in a tax for the bills.

ebond07 (approved):

drallo22 (approved):

PR #594, opened 2023-10-24
Feat/bill cpc 903 create domain client layer

ebond07:
JIRA: https://champlainsaintlambert.atlassian.net/jira/software/c/projects/CPC/boards/35?selectedIssue=CPC-903&sprints=315
## Context:
Added the domain client layer for future use and also made changes to the frontend so that details from other services could be used when looking bills.
## Changes

- When Owner Details is clicked it now directs you to that owner and not like before where it brought you to a blank owner details page.
- When Bill Details is clicked the user can now see the bill ID and the customer's name.
- Created tests to cover the domain client layer

### Owner Details Before
Screenshot 2023-10-23 130053
### Owner Details After
Screenshot 2023-10-23 130532
### Bill Details Before
image
### Bill Details After
Screenshot 2023-10-23 130623

### DIFF
--- a/api-gateway/src/main/resources/static/scripts/bill-history/bill-history.controller.js
+++ b/api-gateway/src/main/resources/static/scripts/bill-history/bill-history.controller.js
@@ -9,25 +9,105 @@ angular.module('billHistory')
         self.unpaidBills = []
         self.overdueBills = []
 
-        // self.owners = [
-        //     { ownerId: '1', firstName: 'George', lastName: 'Franklin' },
-        //     { ownerId: '2', firstName: 'Betty', lastName: 'Davis' },
-        //     { ownerId: '3', firstName: 'Eduardo', lastName: 'Rodriguez' },
-        //     { ownerId: '4', firstName: 'Harold', lastName: 'Davis' },
-        //     { ownerId: '5', firstName: 'Peter', lastName: 'McTavish' },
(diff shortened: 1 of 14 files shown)


Nmazza1 (approved):
I'm assuming by clicking on Owner Details was a button from the Billing section? Otherwise looks good, did not let a single % slip through tests apparently.

api-gateway/src/main/resources/static/scripts/bill-history/bill-history.controller.js
```diff
@@ -22,12 +23,102 @@ angular.module('billHistory')
             { ownerId: '10', firstName: 'Carlos', lastName: 'Esteban' }
         ];
 
+        self.ownersUUID = [
```
drallo22, line 26: Did you try to find a way to obtain the owners without having to hardcode them?

drallo22 (approved):
This is very useful and is a great contribution especially to help those in the next years!

api-gateway/src/main/resources/static/scripts/bill-history/bill-history.controller.js
ebond07: I did and could after several attempts I fell back on this.

Nmazza1: There should be a getAllOwners method, possible implementation for the future year for sure.

PR #599, opened 2023-10-24
feat/BILL-CPC-905_Implement_Pagination_System_For_BillHistory_Page

dazhuox:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-905?atlOrigin=eyJpIjoiNmI1ODdhOWRiM2M4NDJkMTkwZjdhN2Q4NjgyMjg0NjMiLCJwIjoiaiJ9

## Context:
This ticket implements the pagination system for the Bill History page (getAllBillsByPage).

## Changes
- There are buttons now to help the user navigate from one set of list (page) of bills to another.
- The url of the page is changed to the new pagination method.
- The url accepts params that allow the selection of how many items users want to see per page.

Before implementation:
image

After implementation:
image

## Dev notes (Optional)
- pull request link


api-gateway/src/main/java/com/petclinic/bffapigateway/domainclientlayer/BillServiceClient.java
```diff
@@ -56,6 +66,102 @@ public Flux<BillResponseDTO> getAllBilling() {
...
+
+        UriComponentsBuilder builder = UriComponentsBuilder.fromUriString(billServiceUrl + "/bills-pagination");
+
+        builder.queryParam("page", page);
+        builder.queryParam("size",size);
+
+        // Add query parameters conditionally if they are not null or empty
+        if (billId != null && !billId.isEmpty()) {
+            builder.queryParam("billId", billId);
+        }
```
drallo22, line 89: Could have been done using switch or if-else statements but this works fine as well!

drallo22 (approved):
Very useful especially if we have a lot of bills, good job!

api-gateway/src/main/java/com/petclinic/bffapigateway/domainclientlayer/BillServiceClient.java
dazhuox: I didn't think about that, it could definitely work. My usual preference are if statements so I'll stick with that

swafit: > Could have been done using switch or if-else statements but this works fine as well!

In this case an if-else is better since you might have more than one query param that are not provided.

ebond07 (approved):

PR #609, opened 2024-08-29
I am lazy

DylanBrass:
JIRA: link to jira ticket
## Context:
## Changes
## Dev notes (Optional)
- pull request link


api-gateway/src/main/resources/application.yml
```diff
@@ -67,6 +67,6 @@ app:
 
 logging:
   level:
-    root: INFO
+    root: DEBUG
```
nic5694, line 70: Is that needed, won't it just clutter the logs?

nic5694 (commented):
Check the comment i added i think it would be important to change the log level back to INFO instead of DEBUG

nic5694 (approved):
Looks good thanks for applying the changes

PR #620, opened 2024-09-11
Feat/visit cpc 1112 add a new exception in case the email is already sent

FelixAllard:
[JIRA: link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-1112)
## Context:
We are adding a new exception before creating the mailing system so that both service can communicate more easily. Did not implement the exception yet. Did is just in preparation for future changes
## Changes
We first added a new exception. It should not affect any services as it was not implemented and is only related to visits.
Also added some documentation on some functions for the sake of clarity as a team. ( Only one person was on the branch )

### DIFF
--- a/visits-service-new/src/main/java/com/petclinic/visits/visitsservicenew/Exceptions/DuplicateReminderException.java
+++ b/visits-service-new/src/main/java/com/petclinic/visits/visitsservicenew/Exceptions/DuplicateReminderException.java
@@ -1,0 +1,10 @@
+package com.petclinic.visits.visitsservicenew.Exceptions;
+
+/**
+ * Will be called when trying to register a Reminder that is already created for an appointment
+ */
+public class DuplicateReminderException extends RuntimeException {
+    public DuplicateReminderException(String message) {
+        super(message);
+    }
+}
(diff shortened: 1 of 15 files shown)


FelixAllard:
Screenshot 2024-09-11 183425
All the Containers are up and running. Also, the front view works no problem. The changes made here are preliminary changes meant for future implementation!

ioannisp03:
lgtm!

visits-service-new/src/main/java/com/petclinic/visits/visitsservicenew/Exceptions/DuplicateReminderException.java
ioannisp03, line 1: Saw this addition.

ioannisp03 (approved):
lgtm

Valthefirst (approved):
lgtm all tests passed and all 27 containers are up

cgerard321:
Why did you add so many comments? Clean code theory specifies that the code should be readable. Comments should only be added when the code can not be made readable.

FelixAllard:
Me and the rest of the team went through all the code together to make sure we did understand the legacy of the previous team. We thought it would be a good idea to add comments to make sure everything is readable for us who just got our hand  on the code. Those comments where made by team consensus and we added them with Javadoc so that it is mostly to just know what a function does remotely without having to navigate to that specific method. It has proven really useful for us to do that because now we have a way clearer understanding of everything we should be doing.

PR #624, opened 2024-09-13
bug/(AUTH-CPC-1123): Redirect users to home page after login

flyindonut:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-1123
## Context:
When a user logs in, they are redirected to the inventories page. This is problematic because if the user isn't an inventory manager or an admin, they will be unauthorized. Also, users should land on the home page when logging in, not the inventories page.
## Changes

- Redirects users to the home page after logging in
- Changed the routing from Inventories to Home


Futurespast (approved):
This would allow the user to be redirected to home and not get the unauthorized error because of the redirect to inventories.

AndrewBadIdea (approved):
Good adjustment

PR #629, opened 2024-09-14
Feat/BILL-CPC-1046) Allows a User to get all their bills (db bug fix included)

Futurespast:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-1046?atlOrigin=eyJpIjoiYmE3MTgzYzlhZWQ4NGQ4YmI4ZTRhMWYwYjgwYTg2YjgiLCJwIjoiaiJ9

## Context:

This ticket creates the main billing page for the customer and allows them to see all their bills (by customer id). This way the billing in the nav bar now redirects them to their billing page and the customer can see all their bills. **Please note, I have accidently fixed and properly configured the database in this branch. Therefore a new branch will not be created despite the separate story.**

## Changes

- creation of a new customer billing page
- a billing table to get and handle all the bills
- bill model in the front end
- modified path.routes, router and AppNavBar to have the frontend access the newly created page
- fixed the mongo db to actually have customer ids and moved it to the proper layer
- added new controller for old v2 angular frontend

## Added to the UI (No changes to original UI)

image

### DIFF
--- a/api-gateway/src/main/java/com/petclinic/bffapigateway/presentationlayer/v2/BillController.java
+++ b/api-gateway/src/main/java/com/petclinic/bffapigateway/presentationlayer/v2/BillController.java
@@ -1,0 +1,29 @@
+package com.petclinic.bffapigateway.presentationlayer.v2;
+
+import com.petclinic.bffapigateway.domainclientlayer.BillServiceClient;
+import com.petclinic.bffapigateway.dtos.Bills.BillResponseDTO;
+import com.petclinic.bffapigateway.utils.Security.Annotations.IsUserSpecific;
+import com.petclinic.bffapigateway.utils.Security.Variables.Roles;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.http.MediaType;
+import org.springframework.validation.annotation.Validated;
+import org.springframework.web.bind.annotation.*;
+import reactor.core.publisher.Flux;
+
+@RestController
+@RequiredArgsConstructor
+@Slf4j
+@RequestMapping("/api/v2/gateway/bills")
+@Validated
+@CrossOrigin(origins = "http://localhost:3000, http://localhost:80")
+public class BillController {
+    private final BillServiceClient billService;
+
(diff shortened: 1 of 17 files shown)


flyindonut (commented):
Why not use the @SecuredEndpoint annotation for your endpoint in the v2 controller? I think you should write tests to cover the implementation of your endpoint. Also in BillsListTable.tsx, you're using the endpoint of the old API gateway as opposed to the new one you created. You are also using fetch instead of the axiosInstance in the shared folder.

The rest looks good to me!

Futurespast:
The tests for the endpoint where already written, they still pass with the new changes. I didn't use secured annotation because the original billing endpoint in controller didn't use it so I thought it would be ok. I have to use fetch because the axiosinstance was being super weird in processing the flux. I had to try a bunch of different ways to get it to work.

flyindonut:
> The tests for the endpoint where already written, they still pass with the new changes. I didn't use secured annotation because the original billing endpoint in controller didn't use it so I thought it would be ok. I have to use fetch because the axiosinstance was being super weird in processing the flux. I had to try a bunch of different ways to get it to work.

Since you created a new v2 endpoint, I think it would be good to test your new implementation. Also I noticed that the URL contains the customer id. Wouldn't it be better to make the route be '/customer/bills'.? Since you are using the logged in customer's id to perform the request you don't really need to show it in the URL. Each customer will see their corresponding bills when accessing the page.

Futurespast:
> > The tests for the endpoint where already written, they still pass with the new changes. I didn't use secured annotation because the original billing endpoint in controller didn't use it so I thought it would be ok. I have to use fetch because the axiosinstance was being super weird in processing the flux. I had to try a bunch of different ways to get it to work.
> 
> Since you created a new v2 endpoint, I think it would be good to test your new implementation. Also I noticed that the URL contains the customer id. Wouldn't it be better to make the route be '/customer/bills'? Since you are using the logged in customer's id to perform the request you don't really need to show it in the URL. Each customer will see their corresponding bills when accessing the page.

Ok thanks. I will makes these changes. Do I close the pull request?

PR #640, opened 2024-09-16
feat(VETS-CPC-1076): fully implemented the getVetbyVetFirstname and LastName

bluement:
JIRA: link to jira ticket
## Context: https://champlainsaintlambert.atlassian.net/jira/software/c/projects/CPC/boards/15/backlog?selectedIssue=CPC-1076&text=vets
i did this ticket to implement a searchbar to be able to search vets
vets service
service impl
controller
both the apigateway service and controller
the frontend
uifrontend

opted to not do a different card for the frontend because a searchbar is necessary to all other members of my team


flyindonut (commented):
What would happen if two or more vets had the same first or last name? Also, you didn't implement the new endpoints in the API gateway v2.

cgerard321:
The PR template was not completed well. You blew away most of the markdown formatting.

PR #651, opened 2024-09-17
added negative test and added invalidinputexception inside the global…

Futurespast:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-1139

## Context:

Had to add a negative test for the new GetAllBillsByCustomerid endpoint in the bills v2 controller so that the test would be fully covered.

This does not make any changes to the .vscode folder

## Changes

- Made changes inside the controller unit test so that the exception handling would work during tests.
- Added InvalidInputException to GlobalExceptionHandler so it could handle the exception.
- Made the negative test for GetAllBillsByCustomerid when it is given an invalid customer id.
- updated VET testing in api gateway to use the InvalidInputException

## Special Note

Currently I have fixed the API Gateway so that the tests don’t fail but it needs to be updated and also VetServiceClient needs to be updated to properly handle exceptions. Also check vets_service to make sure it's returning the correct statuses

Relevant ticket on Jira:  https://champlainsaintlambert.atlassian.net/browse/CPC-1141

### DIFF
--- a/api-gateway/src/main/java/com/petclinic/bffapigateway/config/GlobalExceptionHandler.java
+++ b/api-gateway/src/main/java/com/petclinic/bffapigateway/config/GlobalExceptionHandler.java
@@ -129,4 +129,11 @@ public class GlobalExceptionHandler {
                 .body(new HttpErrorInfo(HttpStatus.FORBIDDEN.value(),ex.getMessage()));
     }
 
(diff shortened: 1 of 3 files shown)


Futurespast:
The failed check is because of VET not having all their tests pass. Everything I did works fine.

kyrusho (approved):
Thank you for updating our tests in the ApiGatewayControllerTest.java

PR #634, opened 2024-09-15
feat/(CUST-CPC-1068): Show all customers

AndrewBadIdea:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-1068
## Context:
This ticket is about being able, as an Admin, to view every customers (owners) and their personal informations in a page 
## Changes
* Created a new endpoint to get all owners in the OwnerController
* Created the api call method to the new endpoint (getAllOwners.ts)
* Created a page where you are able to see all the customers
* Adding the routing for getting all the customers in "router.tsx"
* Made a test for the get all to verify the users are Admin
## Before and After UI (Required for UI-impacting PRs)
getallcustomers

PR #632, opened 2024-09-14
feat/(CUST-CPC-1075): Being Able To Add Customer To Website

NotZagreus:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-1075
## Context:
Added a v2 endpoint for being able to add a customer (as an admin). Adding a web form to input customer details, following the "Customer Profile Edit" form created previously.
## Changes
- Added the "addOwner" v2 method to the CustomersServiceClient impoving the already existing "createOwner" method, since that one was not operational.
- Created the endpoint to add an owner in the OwnerController
- Created the TypeScript Model for "Owner"
- Created the api call method to the new endpoint (addOwner.ts)
- Based on the "Customer Profile Edit" form, created "Add Customer" form to conform to the addition of a customer.
- Adding the routing for adding the customer in "router.tsx"
- Fixing a small navigation issue in "Profile Edit" 
- Adding the path routes for the form created
- Tested the "addOwner" method 

## Before and After UI (Required for UI-impacting PRs)
image

PR #622, opened 2024-09-13
feat/(CUST-CPC-1077): Update customer information on website

flyindonut:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-1077
## Context:
Created a v2 endpoint to update an owner's information and web page where the logged in customer can update their information.
## Changes
- Created the OwnerController class in the v2 folder of the API gateway
- Created the endpoint to update an owner's information in the Controller
- Created two controller integration tests to cover the implementation of the endpoint
- Added the mockserver depency in build.gradle
- Created the ProfileEdit page where customers can update their information on a pre-filled form. The pre-filled information is based on the logged in customer.
- Created the necessary files for the implementation of the ProfileEdit page (models, API calls, CSS)
## Before and After UI (Required for UI-impacting PRs)
image

PR #656, opened 2024-09-18
feat/(CUST-CPC-1134): Render pages based on user roles

flyindonut:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-1134
## Context:
Currently, users can access every page that the axiosErrorResponseHandler doesn't handle. This means that an admin or vet is able to access the customer profile edit page which only a customer should be able to see.
Absolutely not. I hope.
## Changes
- Created a hook to validate user roles (useHasRequiredRole.ts). Implemented it in the ProtectedRouteProps.tsx
- Added the roles that are authorized to access ProfileEdit, AddingCustomer, and AllOwners. Other teams can implement it to their pages as well (see dev notes).
- Created 4 additional functions that people can use in the UserContext.tsx
- Refactored and reorganized the ProfileEdit.tsx. The update form is now a component in the newly created components folder in features/customers
No change to the UI. Users will just be redirected to 403 Forbidden when they access pages they shouldn't see
## Dev notes (Optional)
If you want to make a page accessible by roles, just take a look at router.tsx. Pages that trigger the axiosErrorResponseHandler will automatically redirect users to an error page. If you want to be consistent, then add roles to each page that would require it.
Update Customer https://github.com/cgerard321/champlain_petclinic/pull/622
Add Customer https://github.com/cgerard321/champlain_petclinic/pull/632
Show all customers https://github.com/cgerard321/champlain_petclinic/pull/634

### DIFF
--- a/petclinic-frontend/src/context/UserContext.tsx
+++ b/petclinic-frontend/src/context/UserContext.tsx
@@ -3,6 +3,7 @@ import { createContext, useContext, useState, ReactNode } from 'react';
 import { UserResponseModel } from '@/shared/models/UserResponseModel';
 import router from '@/router';
 import { AppRoutePaths } from '@/shared/models/path.routes.ts';
+import { Role } from '@/shared/models/Role.ts';
 
 interface UserContextType {
   user: UserResponseModel;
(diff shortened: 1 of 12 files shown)


AndrewBadIdea (approved):
Good fix !

NotZagreus (approved):
Looks good to me

Zako563 (approved):
looks good to me

PR #657, opened 2024-09-18
Implemented the color green/yellow/red based on the status type

vinivelozo:
**JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-1062
## Context:
The ticket is about to be able to see a specific color of the status based on their availability. We are doing this change to facilitate the visual and also it is another way to inform the user to perform a specific task based on the status.
No
## Changes
-The status changes color based on the quantity. 
-The status became red when the quantity is 0, yellow when the quantity less than 20 and green when the quantity is more than 20.
Before
image
After
image


igouriny (approved):
Looks good! .vscode is untouched.

Sunveerg (approved):
Looks good

PR #692, opened 2024-09-19
feat/(PROD-CPC-1120): Filter Products By Sale Price

ThomasBedard:
**JIRA:** (https://champlainsaintlambert.atlassian.net/browse/CPC-1120?atlOrigin=eyJpIjoiYmM4OWZjYmNmM2NkNDhmN2FlZWQ0MTFlYzZjZDg0ZjEiLCJwIjoiaiJ9)
## Context:
I'm adding Filter by Price to the products page since the user wants to be able to shop based on their budget.
This PR doesn't change the .vscode folder

## Changes
This PR adds on top of the getAllProducts to include parameters min and max prices.
It adds tests of the filter feature in the repo testing class
This PR also changes ProductsList.tsx and getAllProducts.ts to accomodate the changes in getAllProducts params.
It changes some aspects of the UI to include the filter min and max prices input fields as well as refining the look of the cards.

## Before and After UI 

Before:
before

After:
after


viktorkuts (commented):
One thing i don't understand is why the change of parsing the Flux in getAllProducts in frontend. It seems more complicated than the way it was before, since \n\n is not always returned and yet is split, the way it was before seems less complicated

ThomasBedard:
> One thing i don't understand is why the change of parsing the Flux in getAllProducts in frontend. It seems more complicated than the way it was before, since \n\n is not always returned and yet is split, the way it was before seems less complicated

you're right I've assumed that the backend would return an SSE (Server-Sent Events). Apparently, this format sends data that is separated by /n/n. I have a version where this doesn't exist i'll push it now.

ZacharyLelievre (dismissed):
The corrections were applied correctly for the GetAllProducts.

Valthefirst:
When I filter for product that's in the range it doesn't show up
image
image
image

Elvis-elvis (approved):
Looks good to me

PR #718, opened 2024-09-21
Feat/INVT CPC 1127 Filter supplies by availability

CharlesS42:
**JIRA:** link to jira ticket (https://champlainsaintlambert.atlassian.net/jira/software/c/projects/CPC/boards/15/backlog?selectedIssue=CPC-1127&text=INVT)
## Context:
Using only the front end, when going into the product list of an inventory, a vet can now search for the products depending on their availability status.
No.

## Changes
Front end only, can now filter the items on its own depending on the product setting passed to it by the fetch function.
Before:
image

After:
- When searching by all statuses
image
- When searching with status with no results
image
- When searching with the proper status
image


Th0masL3 (approved):
Changes looks good and are working.

Sunveerg (approved):
Looks good

PR #743, opened 2024-09-23
bug(VETS-CPC-1141_Make_Service_Client_Handle_Exceptions): made exception handling for client

Apoleh:
**JIRA:** link to jira ticket
## Context:
This ticket is about adding exception handling for client exceptions and specifying the other exception handlings since it was not being done before. Api gateway testing has also been updated accordingly.
No, it does not.

## Changes

- added client exception handling for every method in the VetsServiceClient class
- specified the error for the server exception handling for every method in the VetsServiceClient class
- updated api gateway accordingly


bluement (approved):
looks good

hristoiv11 (approved):
Looks good

PR #760, opened 2024-09-23
Feat/vist cpc 1064 add an emailing system

FelixAllard:
[**JIRA:** link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-1064)
## Context:
We are adding the emailing service within the project. It is wired through the Api gateway and currently support the following feature :

- Create Email Template
- Send email
- See past email sent

## Changes
- Added a new implementation to the Api gateway for the emailing service
- Added a front end to display the sent message and look at which got sent and which didn't
- Before and After UI (Required for UI-impacting PRs)

- The emailing section is new so no before!
Screenshot 2024-09-23 085158
## Dev notes (Optional)
- We are going to add the future email option very soon
- Test are at 82% for the project and 95% for the solution
- There is a little code tangle that doesn't affect workflow but prevented further testing which I should be fixing soon
- IMPORTANT : NEW mailer.env NEEDED FROM NOW ON FOR THE SERVICE TO START


FelixAllard:
We have some Tsx prettier problem, i will modify it and Push once again!

Zako563 (approved):
the result looks good and it does not seem to affect any of us

VladLoghin (approved):

VladLoghin:
very good lots of features

PR #763, opened 2024-09-23
Feat/prod cpc 1030 add search by product name front end3001

J91-cloud:
**JIRA:** : https://champlainsaintlambert.atlassian.net/browse/CPC-1030

## Context:
This feature allows people to search for a specific product by its name. It is not case sensitive. When clikcing on the name of the product that pops up it will display all of the products information. 

This feature does not enteract with the .vscode folder in the front-end. 

## Changes

Added another component and css file that should search for a specific product. I then created an instance of the component inside our products page.

Screenshot 2024-09-23 112643


ThomasBedard (approved):
Works, approving the PR

PR #767, opened 2024-09-23
feat/(PROD-CPC-1037): Recently Seen Products

ZacharyLelievre:
[**JIRA:** link to jira ticket
](https://champlainsaintlambert.atlassian.net/browse/CPC-1037)
## Context:
This ticket focused on implementing a 'Recently Clicked' feature that displays each product selected by the user at the bottom of the page. The feature tracks the products a user interacts with during their session and dynamically updates a section to showcase those products.
The PR does not affect the .vscode

## Changes

- State management: Added useState in ProductList to track the list of recently clicked products.
- Functionality: Implemented a function that updates the list of clicked products and ensures it displays properly on the page.

Before:
image

After:
image

## Dev notes (Optional)
Pull request links


Valthefirst:
When you tested your feature before pushing did you check that it intergrated well with the other features. I don't see the trending section.

J91-cloud (changes requested):
Zachary, It is well done but i recommend that when clicking on new trending products the list is in chronological order. 

For example: 

Screenshot 2024-09-25 154228

Horse Saddle is the last one i clicked but it appears to be first in the list. When i click on a new item it would be added to the right. Next to Bird Cage on the right. 

Maybe you can revert the list.

viktorkuts:
Looks good, but is there a reason why you create your own implementation for the Product when you could've re-used the Product component which is utilized in ProductList?

ThomasBedard:
Screenshot 2024-09-25 at 3 50 41 PM
Screenshot 2024-09-25 at 3 50 49 PM
Screenshot 2024-09-25 at 3 50 59 PM

I'm having multiple issue : 
An added product can't be added to recently seen products
When adding a product to recently view, the clicked card changes styles

J91-cloud (commented):
Zachary, i recommend to maybe enhance this feature. Right now it seems that we are getting the recently clicked products for everyone. It seems to work, because we are testing the web application locally but when this a publicized application  all users will be interfering with this feature. 

Maybe we you can add a endpoint in the back-end that gets all recently clicked with a userId in the URI. This would make it so everyone has its own recently seen product. rather then everyone. 

Please advise me if i am understanding this correctly.

petclinic-frontend/src/features/products/ProductsList.tsx
```diff
@@ -82,6 +85,26 @@ export default function ProductList(): JSX.Element {
...
 
+  const handleProductClick = (product: ProductModel): void => {
+    setRecentlyClickedProducts(listOfProducts => {
+      const updatedProducts = [];
+
+      for (const p of listOfProducts) {
+        if (p.productId !== product.productId) {
+          updatedProducts.push(p);
+        }
+      }
```
ThomasBedard, line 96:
```suggestion
      for (const currentProduct of listOfProducts) {
        if (currentProduct.productId !== product.productId) {
          updatedProducts.push(currentProduct);
        }
      }
```

maybe consider making it more readable with changing "p" to currentProduct

Elvis-elvis (approved):
It looks good but the recent clicked product appears last in the list while I think it should appear first otherwise I like it

Valthefirst:
Why when I refresh the list dissapears?

ZacharyLelievre:
> Looks good, but is there a reason why you create your own implementation for the Product when you could've re-used the Product component which is utilized in ProductList?

You're absolutely right! I hadn't thought of doing it that way. It's now implemented, and I also realized the product ratings were missing from the clicked items, but that's been fixed as well.

ZacharyLelievre:
> Zachary, It is well done but i recommend that when clicking on new trending products the list is in chronological order.
> 
> For example:
> 
> Screenshot 2024-09-25 154228
> 
> Horse Saddle is the last one i clicked but it appears to be first in the list. When i click on a new item it would be added to the right. Next to Bird Cage on the right.
> 
> Maybe you can revert the list.

Thank you for the suggestion! I’ve updated the implementation so that new products are now added in chronological order. Initially, I was using .push(), which added the most recent product to the end of the list. I’ve switched to using .unshift() so the newest clicked item now appears at the start of the list, making the most recent clicked product always displayed first.

J91-cloud (dismissed):

ZacharyLelievre:
> When you tested your feature before pushing did you check that it intergrated well with the other features. I don't see the trending section and the search bar in the screenshots.

I didn’t include the top section in the original screenshot, but here it is with the feature working:
image

Regarding the issue you mentioned about refreshing the page, I’ve resolved it by using localstorage to store each clicked product on the user’s computer.

As for the owner issue, I haven’t encountered that problem on my end. It seems to be working correctly for all roles. Let me know if you’re still experiencing it.

ZacharyLelievre:
> Do you know why your feature doesn't work when I login as an owner?

As for the owner issue, I haven’t encountered that problem on my end. It seems to be working correctly for all roles. Let me know if you’re still experiencing it.

petclinic-frontend/src/features/products/ProductsList.tsx
ZacharyLelievre: Good point! I've updated p to currentProduct to make the code more readable. Thanks for the suggestion!

ZacharyLelievre:
> Zachary, i recommend to maybe enhance this feature. Right now it seems that we are getting the recently clicked products for everyone. It seems to work, because we are testing the web application locally but when this a publicized application all users will be interfering with this feature.
> 
> Maybe we you can add a endpoint in the back-end that gets all recently clicked with a userId in the URI. This would make it so everyone has its own recently seen product. rather then everyone.
> 
> Please advise me if i am understanding this correctly.

Actually, since I'm using localStorage, each user will already have their own recently clicked products stored locally, so this issue should be resolved. We can still look into adding an endpoint in the future if needed for next Sprint!

ZacharyLelievre:
> Screenshot 2024-09-25 at 3 50 41 PM Screenshot 2024-09-25 at 3 50 49 PM Screenshot 2024-09-25 at 3 50 59 PM
> I'm having multiple issues : An added product can't be added to recently seen products When adding a product to recently view, the clicked card changes styles

That’s a good thing you pointed out, but I just wanted to clarify that the issue you're experiencing is actually due to how the requestCount function is working. When a product is added and clicked, the requestCount gets triggered, but it’s initially set to null, which causes a 500 error. This isn't a problem with this PR itself but rather with how the requestCount handles null values which I did not touch.

J91-cloud (approved):
Great! Thanks for fixing the order of which we see the recently clicked products.

ThomasBedard (approved):
Fixes were applied, works well 👍

Valthefirst (approved):

PR #779, opened 2024-09-26
Implemented a delete button for delete a supply

vinivelozo:
**JIRA:https://champlainsaintlambert.atlassian.net/browse/CPC-1152
## Context:
-Implement a delete button for deleting supplies inside an inventory
## Changes
-Added a delete button inside InventoryProducts.tsx
-When the delete button is pressed, the supply is deleted in the database as well
Before
image
After
image
## Dev notes (Optional)
-Only changed the front end since the back end and all test are already done

### DIFF
--- a/petclinic-frontend/src/features/inventories/InventoryProducts.tsx
+++ b/petclinic-frontend/src/features/inventories/InventoryProducts.tsx
@@ -28,6 +28,18 @@ const InventoryProducts: React.FC = () => {
       setLoading(false);
     }
   };
+  // Delete product by productId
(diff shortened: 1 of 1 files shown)


Carlos-123321 (approved):
The deleted functionality for the supplies was something we really needed. Maybe in the future we could start thinking about ways to style it to cater it to our user's experience but for the moment it is a great initiative.

CharlesS42 (approved):
The buttons are displaying as it should, so great job.
(nitpick) You could replace the delete button by a trash icon.

igouriny:
The PR doesn't follow proper naming conventions, consider changing it.

PR #780, opened 2024-09-26
feat/PROD-CPC-1035_Filtering_Products_By_Products_Type

Elvis-elvis:
**JIRA:** link to jira ticket
## Context:
Implementing a filter feature to help guests search for products easily

## Changes
- Implemented filtering functionality to allow users to filter products by type and price.
- Updated the ProductList component to include error handling for filter inputs.
- Modified the AddProduct component to ensure that new products can be added successfully after filtering.
- Adjusted relevant API calls in the backend to support new filtering options..
Screenshot 2024-09-23 005235

## Dev notes (Optional)
Screenshot 2024-09-23 005641
Screenshot 2024-09-23 005641
Screenshot 2024-09-23 005727

Pull request links


docker-compose.yml
```diff
@@ -465,4 +465,4 @@ services:
       - MYSQL_DATABASE=MyAppDb
       - SPRING_PROFILES_ACTIVE=docker
     ports:
-      - "3306:3306"
+      - "3309:3306"
```
ThomasBedard, line 468: I'm curious on why this was changed

ThomasBedard:
I'm experiencing some problems with the implementation of filter products by type. 
Screenshot 2024-09-26 at 10 02 21 AM
**The main problem** is that I had to go in the code to see what the accepted types were for products. I think a drop down menu with the product type would work better.
**Another issue** is that i can't clear the product by type field with the clear filters button. If you implement the drop down menu as mentioned above, you can just add your useState set method to clearFilters in ProductsList.tsx and set it to an empty string or something like that.
The last problem is that i can't filter by price if the filter by product type is active.

products-service/src/main/java/com/petclinic/products/presentationlayer/products/ProductController.java
```diff
@@ -68,5 +68,9 @@ public Mono<ResponseEntity<ProductResponseModel>> deleteProduct(@PathVariable St
                 .map(ResponseEntity::ok)
                 .defaultIfEmpty(ResponseEntity.badRequest().build());
     }
+    @GetMapping(value = "/filter/{productType}", produces = MediaType.APPLICATION_JSON_VALUE)
```
J91-cloud, line 71: We implemented GetALL's using a stream value when using flux. Have you tried implemented this?

J91-cloud (commented):
I recommend you implement testing for this feature. Our service will most likely lose test coverage because the feature is adding methods in serviceImpl and controller class. 

As well as in the api-gateway

docker-compose.yml
Elvis-elvis: I had changed it because git bash when I was doing docker compose up it would tell me I had a problem with my ports and somebody else in the class had that same issue and they told me to change the port, and it worked, but I forgot to change it back to 3306 when I was doing the push. Thanks for letting me know

Elvis-elvis:
> I'm experiencing some problems with the implementation of filter products by type. Screenshot 2024-09-26 at 10 02 21 AM
> 
> ### **The main problem**
> I had to go in the code to see what the accepted types were for products. I think a drop down menu with the product type would work better.
> 
> ### **Another issue**
> I can't clear the product by type field with the clear filters button. If you implement the drop down menu as mentioned above, you can just add your useState set method to clearFilters in ProductsList.tsx and set it to an empty string or something like that.
> 
> ### The last problem
> I can't filter by price if the filter by product type is active.

Thanks on the review Thomas. I'll take a look and see how I can make both our features work together aswell as the clear button

ThomasBedard (approved):
The docker compose yml change has been reverted and the main "issue" will be fixed in another PR from elvis

ZacharyLelievre (approved):
The issues identified during the review have been resolved: the missing tests were implemented in the product service, and the port number switch was fixed. I’ll go ahead and approve the review, but it might be worth considering using the existing getAll method with a stream value instead of creating a new one.

PR #787, opened 2024-09-27
feat(CUST-CPC-1190): Admin/Vets can select an owner and view their information

AndrewBadIdea:
… owner: Andrew Badea

**JIRA:** https://champlainsaintlambert.atlassian.net/jira/software/c/projects/CPC/boards/15/backlog?epics=visible&selectedIssue=CPC-1190
## Context:
When the admin or vet clicks on the id link it goes to a new page where it displays all the owner's information
NO
## Changes
* Deleted the getbyid code in the AllOwners.tsx (moved some parts in the new page)
* updated the css
* Owner's Personal information
* Owner's pet information
* Owner's bill information
* Added a delete button in the new page
* Added go back button
Before
image
filter

After
image
image


NotZagreus (commented):
It looks great and all, my question is why is there like +3713 and -3696 (lines I'm guessing), the package-lock.json doesn't show for me lol (it's just a question anyway)

flyindonut (approved):
Everything looks good. The only suggestion I have is that since you are using axios, I think it would be good to use the shared axios instance.

petclinic-frontend/src/features/customers/components/CustomerDetails.css
```diff
@@ -1,42 +1,80 @@
...
+
+.section h3 {
+  color: #007bff;
+  margin-bottom: 15px;
   font-size: 1.5em;
 }
 
-.customer-details-card p {
-  margin: 12px 0;
+.section p, .section ul {
```
Justin222993, line 45: NITPICK: Although really not important, I think you should still be using classes such as customer-details-section to make sure only the css you want to affect is being affected

Justin222993 (commented):
It looks good, however, I would add the pet ID in the table since we might want to be able to click on it as a link and modify the pet of an owner easily and dynamically. If the id is missing, it is a lot harder to find the pet in question and modify it from the front end. Maybe as another story you could add the fact that there's clickable pet links that bring you to their page and you can modify them there as well as see the information.

Justin222993 (approved):

AndrewBadIdea:
> It looks great and all, my question is why is there like +3713 and -3696 (lines I'm guessing), the package-lock.json doesn't show for me lol (it's just a question anyway)

On my way to fix that :)

AndrewBadIdea:
> It looks good, however, I would add the pet ID in the table since we might want to be able to click on it as a link and modify the pet of an owner easily and dynamically. If the id is missing, it is a lot harder to find the pet in question and modify it from the front end. Maybe as another story you could add the fact that there's clickable pet links that bring you to their page and you can modify them there as well as see the information.

I will definetly do that since my full stack story is about updating pet

PR #790, opened 2024-09-28
bug(VETS-CPC-1248): Fix delete vet by vet ID

orell2j:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-1248?atlOrigin=eyJpIjoiYmRmMDM1NmMyNjE4NGY1ZmE0OGJlMDQxZWQ5NzA2OWYiLCJwIjoiaiJ9

## Context:

This ticket is about deleting a veterinarian by their ID, the admin needs to have this ability to ensure that the list of vets is up-to-date, especially when a vet can quit or be fired. 

So this change fixes a problem where the vet was not properly deleted from the database. Also, it fixes a bug where after clicking the "delete" button, it was navigating the admin to the vet's detail page instead of staying on the list.

No, this PR does not modify the .vscode folder

## Changes

- Fixed the delete functionality so that it properly deletes the veterinarian record from database (MongoDB).

- Addressed a bug where, after deleting a vet, the UI would navigate to the vet's detail page. Now, the page stays on the veterinarian list and removes the deleted vet from the UI.

BEFORE:

1- Initial UI with "John Doe":

image

2- Attempting to Delete "John Doe": 

image

3- Reloading the page still shows the deleted vet: 

image

AFTER:

1- Initial UI with "John Doe": 

image

2- Approval to delete this vet:

image

3- Deleted "John Doe" from the UI:

image

4- MongoDB reflecting the deletion:

image

## Dev notes (Optional)
- this fix ensures the deletion in the backend and after deletion the vet list is updated in real-time without refreshing the page

### DIFF
--- a/petclinic-frontend/src/features/veterinarians/VetListTable.tsx
+++ b/petclinic-frontend/src/features/veterinarians/VetListTable.tsx
@@ -44,7 +44,10 @@ export default function VetListTable(): JSX.Element {
     navigate(`/vets/${vetId}`);
   };
 
-  const handleVetDelete = async (event: React.MouseEvent, vetId: string): Promise<void> => {
+  const handleVetDelete = async (
+    event: React.MouseEvent,
+    vetId: string
+  ): Promise<void> => {
     event.stopPropagation();
     try {
       await deleteVet(vetId);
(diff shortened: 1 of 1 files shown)


Edwxd (approved):
The code looks better. No major file changes like the last PR which is good.  The handler for the delete is implemented better then before. Overall good job.

hristoiv11 (approved):
The code looks better. It's good that you fixed this problem and that you showed that it is deleted from the database.

PR #791, opened 2024-09-28
bug(VETS-CPC-1279): bug fix for add vet

Edwxd:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-1276

## Context:
After adding a vet, we had to manually reload the page to be able to see the newly added vet. So I added an automatic reload for the page once a new vet was added. I also realised that the form for the add did not match the one for the update. So I decided to change them both to set all fields in a logical order. 

This PR does not effect the .vscode file.

## Changes

- Added an automatic reload to the page once a new vet was added.
- Re-organized the form for add and update so that they are the same and fields are in a logical order.

Before: 

- Page didn't reload
- Fields in the forms where not in the same order. 

After:

Field in add and update form are in the same order:
image
image

Pages relaods on its won after adding a new vet.

## Dev notes (Optional)
Pull request links


hristoiv11:
Can you add the before and after UI of the page after the add?

Apoleh:
Your PR should be: **bug(VETS-CPC-1279): bug fix for add vet** since you cannot use capitals or punctuation. Apart from that and not having a 'before UI' picture, everything else looks good.

hristoiv11 (approved):
I suggest putting UI after every change, in this case for the page reloading, this way would be easier to see if it worked. Apart from this, Good work.

kyrusho (approved):
Great job Edward, everything seems good.

PR #796, opened 2024-09-29
Feat(CUST-CPC-1188): Users can register on the website using a sign up page

Justin222993:
[**JIRA:** link to jira ticket](https://champlainsaintlambert.atlassian.net/jira/software/c/projects/CPC/boards/15/backlog?selectedIssue=CPC-1188)
## Context:
It doesn't do any changes to the vscode folder

## Changes
- Added sign up page and relevant css
- Created V2 functions in the api gateway for createUser
- Created V2 functions in the api gateway for verifyUser
- Added EmailSendingFailedException
- Added 3 tests in new class v2/AuthControllerUnitTest
- Added 3 tests in new class v2/domainclientlayer/AuthServiceClientControllerIntegrationTest
- Added 1 test in new class businesslayer/AuthServiceUserServiceTests
- Changed from using the old emailing service to the new one to send verification email
- Removed a cart post api request that was added in the UserServiceImpl in auth service as the person implementing added it and pulled without testing and it never functioned, so it was breaking sign up. I have contacted and will be fixing this problem in a future pull request.
- Added Models in the backend to communicate with the new emailing service  (DirectEmailModelRequestDTO and EmailModelResponseDTO) as well as added the domain client for it
- Added Models needed in the front end to send the user and customer details to the backend (RegisterModel and OwnerRequestModel)
- Changed the redirection of the verification link to the new front end after the token is verified
- Changed the relevant links on the website to go to the sign up page
- Made the page reload on logout so it deletes user info and you can't just press back and be logged in again
- temporarily reduced jacoco testing minimum to 85 in auth service
Before there was no sign up page, now:
image
image

It also has instant validation, if you modify a field like write something and remove it, it will tell you a field is required, and as you write your password it will tell you what is required. Max limit of character for everything but email: 60, max for email: 320, Phone needs to be an uninterrupted 10 numbers, email is checking for the email format and password is checking for "Password must contain at least 8 characters, including an uppercase letter, a lowercase letter, a number, and a special character."
image

Loading:
image

Here is a successful email being sent:
On the user's end:
image

On the emailing end:
image

## Dev notes
<span style="color:red">**IF YOU ARE USING AN ANTIVIRUS**</span>
The antivirus might be blocking your access to the mailing API, if the email is not being sent after submitting the form, please make sure to either grant access to the api in your antivirus (The SMTP server) or disable it while you are trying to sign up.

The auth-service minimum jacoco testing has been temporarily lowered to 85, I am waiting for another PR to be able to do the remaining tests

### DIFF
--- a/api-gateway/src/main/java/com/petclinic/bffapigateway/domainclientlayer/AuthServiceClient.java
+++ b/api-gateway/src/main/java/com/petclinic/bffapigateway/domainclientlayer/AuthServiceClient.java
@@ -138,7 +138,6 @@ public class AuthServiceClient {
 
     }
 
-    //STRINGTOFINDNEWMETHODSTODOTESTINON-GEIFUSEFUISEB
     public Mono<OwnerResponseDTO> createUserUsingV2Endpoint(Mono<Register> model) {
 
         String uuid = UUID.randomUUID().toString();
@@ -271,7 +270,6 @@ public class AuthServiceClient {
                 });
     }
 
-    //STRINGTOFINDNEWMETHODSTODOTESTINON-GEIFUSEFUISEB
     public Mono<ResponseEntity<UserDetails>> verifyUserUsingV2Endpoint(final String token) {
 
         return webClientBuilder.build()
--- a/api-gateway/src/main/java/com/petclinic/bffapigateway/domainclientlayer/AuthServiceClient.java
+++ b/api-gateway/src/main/java/com/petclinic/bffapigateway/domainclientlayer/AuthServiceClient.java
@@ -138,6 +138,45 @@ public class AuthServiceClient {
 
     }
 
+    //STRINGTOFINDNEWMETHODSTODOTESTINON-GEIFUSEFUISEB
+    public Mono<OwnerResponseDTO> createUserUsingV2Endpoint(Mono<Register> model) {
+
+        String uuid = UUID.randomUUID().toString();
+
+        return model.flatMap(register -> {
+                    register.setUserId(uuid);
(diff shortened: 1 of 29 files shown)


AndrewBadIdea (dismissed):
Great work on this! Everything looks well-structured and efficient

AndrewBadIdea (dismissed):
Great work on this! Everything looks well-structured and efficient

Justin222993:
> The logging you added to the emailing-service should be removed so it does not print out in the logs confidential data

Thank you, it has been fixed!

flyindonut (approved):
Great work! Just wondering why not use the shared axios instance and the new endpoint in the request in signup.tsx. Everything else looks good to me.

AndrewBadIdea (dismissed):

flyindonut (dismissed):

AndrewBadIdea (approved):

flyindonut (approved):

Justin222993:
> Great work! Just wondering why not use the shared axios instance and the new endpoint in the request in signup.tsx. Everything else looks good to me.

You're right, I didn't use the axios instance! But I noticed that a lot of pages are also not using it. We should do a bug fix where we change all of them to the axios instance.

cgerard321:
> > Great work! Just wondering why not use the shared axios instance and the new endpoint in the request in signup.tsx. Everything else looks good to me.
> 
> You're right, I didn't use the axios instance! But I noticed that a lot of pages are also not using it. We should do a bug fix where we change all of them to the axios instance.

Please go ahead and enter this issue into Jira.

PR #810, opened 2024-10-01
Bug(CUST-CPC-1188): Fixed V2 endpoint for creating a user failing and reverted back to old emailing service

Justin222993:
[**JIRA:** link to jira ticket](https://champlainsaintlambert.atlassian.net/jira/software/c/projects/CPC/boards/15/backlog?selectedIssue=CPC-1188)
## Context:
We are doing this change because Christine requested the change to the older mailing service and because there was a small bug with the v2 endpoint for creating an owner account.
It doesn't do any changes to the vscode folder

## Changes
- Reverted back to the old emailing service at Christine's request
- Fixed a small bug with the v2 endpoint for creating a customer as it was overwriting the UUID
- Upped Jacoco minimum % to 90
- Changed to v2 endpoint since the bug was fixed

Didn't Impact UI

## Dev notes
There should no longer be any problems caused by antiviruses or limits.

### DIFF
--- a/api-gateway/src/main/java/com/petclinic/bffapigateway/domainclientlayer/CustomersServiceClient.java
+++ b/api-gateway/src/main/java/com/petclinic/bffapigateway/domainclientlayer/CustomersServiceClient.java
@@ -162,16 +162,12 @@ public class CustomersServiceClient {
     }
 
     public Mono<OwnerResponseDTO> addOwner(Mono<OwnerRequestDTO> model) {
-        String ownerId = UUID.randomUUID().toString();
-        return model.flatMap(requestDTO -> {
-            requestDTO.setOwnerId(ownerId);
-            return webClientBuilder.build()
-                    .post()
-                    .uri(customersServiceUrl + "/owners")
-                    .body(BodyInserters.fromValue(requestDTO))
-                    .retrieve()
-                    .bodyToMono(OwnerResponseDTO.class);
-        });
+        return model.flatMap(requestDTO -> webClientBuilder.build()
+                .post()
+                .uri(customersServiceUrl + "/owners")
+                .body(BodyInserters.fromValue(requestDTO))
+                .retrieve()
+                .bodyToMono(OwnerResponseDTO.class));
     }
 
     public Flux<PetType> getPetTypes() {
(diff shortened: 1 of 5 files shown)


auth-service/build.gradle
NotZagreus, line 1: How did we not see this before lol

api-gateway/src/main/java/com/petclinic/bffapigateway/domainclientlayer/CustomersServiceClient.java
```diff
@@ -162,16 +162,12 @@ public Mono<OwnerResponseDTO> createOwner(OwnerRequestDTO model) {
...
-                    .body(BodyInserters.fromValue(requestDTO))
-                    .retrieve()
-                    .bodyToMono(OwnerResponseDTO.class);
-        });
+        return model.flatMap(requestDTO -> webClientBuilder.build()
+                .post()
+                .uri(customersServiceUrl + "/owners")
+                .body(BodyInserters.fromValue(requestDTO))
+                .retrieve()
+                .bodyToMono(OwnerResponseDTO.class));
```
NotZagreus, line 170: Thank you I did this and I see what the problem is thanks for removing the redundant UUID generation for ownerId and explaining it to me

NotZagreus (approved):
Good changes especially catching my mistake in the CustomersServiceClient. Thanks Justin!

auth-service/build.gradle
Justin222993: If you're referring to the resetting of the UUID in the new V2 endpoint, it's because the sign-up page wasn't done yet and the profile page, which dependent on a version of the UUID also wasn't made yet. So the resetting of the UUID didn't affect much and was hard to find until I made the sign up page and compared the V1 to V2 endpoints. (Since V1 was working perfectly)

AndrewBadIdea (approved):

api-gateway/src/main/java/com/petclinic/bffapigateway/domainclientlayer/CustomersServiceClient.java
Justin222993: No problem!

PR #816, opened 2024-10-02
Feat/prod cpc 1178 products detailed page

Elvis-elvis:
**JIRA:** link to jira ticket
## Context:
As a Pet Owner, I want to view the detailed information of a product so that I can make an informed decision before purchasing it.

## Changes
- Implemented a Product Details page that displays detailed information about selected products.
- Updated the routing to navigate to the Product Details page from the product list.
- Enhanced the product model to include additional properties (e.g., description, brand, dimensions).
- Integrated the backend API to fetch product details dynamically.
- Added error handling for scenarios where product details cannot be retrieved.
image
image

## Dev notes (Optional)
Pull request links


petclinic-frontend/src/features/products/components/Product.tsx
```diff
@@ -51,14 +59,14 @@ function Product({ product }: { product: ProductModel }): JSX.Element {
...
-    }
-  };
+  // const handleProductClick = async (productId: string): Promise<void> => {
+  //   try {
+  //     const product = await getProductByProductId(productId);
+  //     setSelectedProduct(product);
+  //   } catch (error) {
+  //     console.error('Failed to fetch product details:', error);
+  //   }
+  // };
```
ThomasBedard, line 69: i think its better to delete this completely as you implemented a new product handle click above

petclinic-frontend/src/features/products/components/Product.tsx
```diff
@@ -14,6 +16,12 @@ function Product({ product }: { product: ProductModel }): JSX.Element {
     null
   );
 
+  const navigate = useNavigate(); // Get the navigate function
+  const handleButtonClick = (): void => {
```
ThomasBedard, line 20: This should be named something like handleProductClick or handleProductTitleClick in my opinion. My reasoning behind this change is that you can only trigger this function when clicking from the title of a product.

ThomasBedard (commented):
It works great ! However, there is a minor issue with the rating as when we select a rating, then click on the product to get more details, it doesn't show the rating until the second time we click on the product. Same thing when deleting the rating, it doesn't update until the second time we click on the product.

Screenshot 2024-10-02 at 2 37 34 PM
Screenshot 2024-10-02 at 2 38 11 PM
Screenshot 2024-10-02 at 2 38 45 PM

petclinic-frontend/src/features/products/api/ProductDetails.tsx
viktorkuts, line 1: Is this just placeholder values if it can't load or is there no productDescription field in the backend?

viktorkuts (commented):
Looks good, just need to refresh the product details when clicking on product (like in Ratings you can check, I do this for the product and starrating, where it fetches everytime it calls an update). Also hardcoded description is OK for now until the productDescription field is added into the backend.

PR #826, opened 2024-10-02
Bug(CUST-CPC-1188): Fixed lost UUID when creating an owner account as an admin

Justin222993:
**JIRA:** link to jira ticket
https://champlainsaintlambert.atlassian.net/jira/software/c/projects/CPC/boards/15/backlog?selectedIssue=CPC-1188
## Context:
My previous pull request broke the add owner and I recently found out so I fixed it
It doesn't go NEAR the .vscode foler :)

## Changes
Changed the CustomersServiceClient to set a UUID if when sending a request it is null or empty. It does not reset it if it is not.
Didn't affect front end, but here was the issue: 
image
It is now fixed


AndrewBadIdea (approved):
Good fix !

AlexCristea04 (approved):
The service client now assures that when processed, there's a valid ownerId. Good!

PR #827, opened 2024-10-02
feat(INVT-CPC-1232): Display Inventory Images

Carlos-123321:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-1232
## Context:
This ticket mainly introduces the implementation of two fields for our **Inventory** entity: 

**inventoryImage** and **inventoryBackupImage**. This ticket let's us display an image for an inventory and in case the inventoryImage is **invalid** it switches the display to be the **backup image**. In case the backup image **also fails**, there is a default image it switches to **in the frontend.** 
**No, it does not**
## Changes
Other notable changes include:

• **Removed Supply Entity**

• **Removed redundant id field** from the Product and ProductResponseDTO class(We already have a productId field)

•  **Removed testing logic** which involved the use of the **id field** and changed it to instead used the productId field

• **Updated testing logic** in inventory-service and api-gateway to **include** the inventoryImage and inventoryBackupImage fields

• **Updated** the Add Inventory and the Edit Inventory **forms** to include labels to add inventory images.

• **Many more changes...**

**Before** this is how the cards looked like:

image

**After** this is how the cards look:

image

**Before** this is how the Add Inventory form looked like:

image

**After** this is how the Add Inventory form looks like:

image

**Before** this is how the Edit Inventory form looked like:

image

**After** this is how the Edit Inventory form looks like:

image

## Dev notes (Optional)
Pull request links


inventory-service/src/main/java/com/petclinic/inventoryservice/datalayer/Product/Product.java
```diff
@@ -11,7 +10,6 @@
 @EqualsAndHashCode
 public class Product {
     @Id
-    private String id;
```
CharlesS42, line 14: Why did you take out the id?

Carlos-123321: So my initial intention was to combine the private id(Our database one) to the public one(The one our users can access) but after further consultation I learned that this might not be the best practice for security purposes. I will fix this and thank you for pointing it out.

PR #832, opened 2024-10-03
Feat/INVT/CPC-1200/Display Quantity Of Products Inside An Inventory

Th0masL3:
[**JIRA:** link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-1200)
## Context:
I'm doing this change to display the quantity of products in each inventory. 
No it doesn't.

## Changes
It display the quantity of products in each inventory. It doesn't affect any other modules. 

Before:
image

After: 

image

### DIFF
--- a/petclinic-frontend/src/features/inventories/InventoriesListTable.tsx
+++ b/petclinic-frontend/src/features/inventories/InventoriesListTable.tsx
@@ -31,6 +31,9 @@ export default function InventoriesListTable(): JSX.Element {
   const [lowStockProductsByInventory, setLowStockProductsByInventory] =
     useState<{ [inventoryName: string]: ProductModel[] }>({});
   const [showLowStock, setShowLowStock] = useState(false);
+  const [productQuantities, setProductQuantities] = useState<{
+    [key: string]: number;
+  }>({});
 
   const toggleAddSupplyModal = (): void => {
     setShowAddSupplyModal(prev => !prev);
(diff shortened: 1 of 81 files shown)


petclinic-frontend/src/features/inventories/InventoriesListTable.tsx
```diff
@@ -94,6 +97,39 @@ export default function InventoriesListTable(): JSX.Element {
     }
   };
 
+  useEffect(() => {
+    if (inventoryList.length > 0) {
+      inventoryList.forEach(inventory => {
+        fetchProductQuantity(inventory.inventoryId);
+      });
+    }
+  }, [inventoryList]);
```
CharlesS42, line 106: I think binding the product quantity of an inventory to the event where the inventoryList would be changed is a good idea. Later, we could also make it so that when a product within an inventory is added or deleted, that the action would notify its inventory that it's amount of products has changed.

CharlesS42 (approved):

petclinic-frontend/src/features/inventories/InventoriesListTable.tsx
Th0masL3: Good idea, thanks for the feedback!

petclinic-frontend/src/features/inventories/InventoriesListTable.tsx
```diff
@@ -94,6 +97,39 @@ export default function InventoriesListTable(): JSX.Element {
...
+
+  const fetchProductQuantity = async (inventoryId: string): Promise<void> => {
+    try {
+      const response = await fetch(
+        `http://localhost:8080/api/v2/gateway/inventories/${inventoryId}/productquantity`,
```
Sunveerg, line 111: Maybe you should create another file to handle promises and fetching

Sunveerg (approved):
Looks good maybe you could separate the fetching and promises in different files.

petclinic-frontend/src/features/inventories/InventoriesListTable.tsx
Th0masL3: I find it easier to do it that way but I'll consider changing it later on. Thanks for the feedback!

PR #842, opened 2024-10-04
Fixed bug, changed card to cardProduct

J91-cloud:
**JIRA:** link: https://champlainsaintlambert.atlassian.net/browse/CPC-1291
## Context:

Fixing a card style bug for the second time after inventory created a additional CSS file that override ours. 

## Changes
I changed a className inside Card.css in inventory service from card to cardProduct. This is for to hrlp understand that this card styling is for us while the other called card is for inventory. 

I then needed to change everything we className we had in our files from card to cardProduct. 

## Before UI

image

## After UI
image

## Dev notes (Optional)
Pull request links


Carlos-123321 (approved):
Sorry about that forgot css styles are globally applied. I converted my css files to modules so it doesn't happen again.

Valthefirst (approved):
It's all good the low stock messages are back now. All the files needed to be modified have been and there's only one author to the PR

viktorkuts (approved):
Don't have much to say, pretty straight forward, css styling changes. Accommodating the out-of-stock check also looks good. Good work!

Carlos-123321 (approved):
Approved

PR #850, opened 2024-10-05
feat(CART-CPC-1273): Remove product from cart

NtsoaElie:
**JIRA:** [link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-1273)
## Context:
When you go in carts and then view carts, you can see cart products. This ticket implements the feature to remove a product in cart.

## Changes
- Addition of removeProductFromCart method
- Implementation of the said method
- Implements tests
- Changed UserCart.tsx for it to be able to handle the deletion of an item present in cart
BEFORE:
image
AFTER: 
image
image

### DIFF
--- a/api-gateway/src/main/java/com/petclinic/bffapigateway/domainclientlayer/CartServiceClient.java
+++ b/api-gateway/src/main/java/com/petclinic/bffapigateway/domainclientlayer/CartServiceClient.java
@@ -44,6 +44,15 @@ public class CartServiceClient {
                 .bodyToMono(CartResponseDTO.class);
     }
 
+    public Mono<CartResponseDTO> removeProductFromCart(String cartId, String productId){
+        return webClientBuilder.build()
+                .delete()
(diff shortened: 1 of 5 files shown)


emailing-service/emailing-service-test/obj/Debug/net8.0/emailing-service-test.GeneratedMSBuildEditorConfig.editorconfig
AlexCristea04, line 1: For future reviewers: I spoke with Felix (he created the emailing service) and I ran it myself to see if it would break. Luckily, every thing is still in order. Next time, avoid directly building the project and use docker instead.

AlexCristea04 (dismissed):
Look over my comment for next time. Every thing seems to be working otherwise though.

AlexCristea04 (approved):

Hares-2088 (approved):
good job

PR #851, opened 2024-10-05
Feat/CUST-CPC-1192 admin deactivate customer account

Jesse-Bourassa:
[**JIRA:** link to jira ticket]
(https://champlainsaintlambert.atlassian.net/browse/CPC-1192?atlOrigin=eyJpIjoiODVlZjMxNmYzODU5NGE3NzllNzE2ZTY0ZWZmMzBiYWIiLCJwIjoiaiJ9)
## Context:
Admin can now disable an account instead of deleting it so users cant have access to there account

NO

## Changes

Added a isDisable boolean to the user Entity to check weather or not the account is disabled
Only admin can disable an account
Added patch Methods for EnableUser and DisableUser
Made test for these methods too
change the front end to have a button to disable the account in customerDetails
added is_disabled in the auth_db as a boolean

Before
image
After
image


api-gateway/src/main/java/com/petclinic/bffapigateway/dtos/Auth/UserDetails.java
```diff
@@ -19,5 +19,6 @@ public class UserDetails {
     private String email;
 
     private Set<Role> roles;
+    private boolean isDisabled;
```
NotZagreus, line 22: I would change the name of this boolean, it's a bit weird, maybe we make it to "isActivated"? or something of sorts. But your logic looks alright, good PR

NotZagreus (approved):
Read my comment and we can change that another time, it's not a pressing matter but the rest is nice.

api-gateway/src/main/java/com/petclinic/bffapigateway/dtos/Auth/UserDetails.java
Jesse-Bourassa: image
I understand your point, but the naming fits with the code as if you look at my method when you setDisable it becomes true, but your right il change it for next sprint

flyindonut (approved):
Great job! I think you should've added a screenshot of the error message when logging in as a disabled user. Here is one:

image

The buttons look a bit wonky compared to the before screenshot, but the implementation looks good to me.

PR #855, opened 2024-10-05
feat(CART-CPC-1256): implemented the wish list and worked on a better ui for the user cart page

Hares-2088:
**JIRA:** [ link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-1256)
## Context:
I implemented a wish list so the user can put products on another list besides his cart.
no it doesn't 

## Changes
- change the entity
- all the DTOs that are concerned 
- UserCart.tsx 
- fixed a problem in the checkout function
- many modifications on UserCart.css and ItemCart.css

## Before UI
image
## After UI
Screenshot 2024-10-05 142708


petclinic-frontend/src/features/carts/components/UserCart.tsx
```diff
@@ -266,57 +269,94 @@ const UserCart = (): JSX.Element => {
   return (
     <div className="UserCart">
       <NavBar />
```
flyindonut, line 271: In case you're wondering why the Navbar in this page looks different from the other pages, it's because it's applying a css class to the Navbar. You could use an extra div and move the Navbar out or you could also create another Page where you import the Navbar component and the UserCart component. I think the second option is the best since everyone is doing that.

api-gateway/src/main/java/com/petclinic/bffapigateway/dtos/Cart/CartResponseDTO.java
AlexCristea04, line 1: Good idea to just make it a list of response models instead of an entirely different entity!

AlexCristea04 (approved):

AlexCristea04 (approved):

cart-service/src/test/java/com/petclinic/cartsservice/presentationlayer/CartControllerUnitTest.java
FelixAllard, line 1: Test well modified

petclinic-frontend/src/features/carts/components/CartItem.css
FelixAllard, line 1: Good changes on the css! I see it with the picture!

FelixAllard (approved):

PR #897, opened 2024-10-07
feat(PROD-CPC-1135): reviews

viktorkuts:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-1135
## Context:
Reviews is an add-on to the already implemented Ratings from Sprint 1.
A user can now leave a text review and see other reviews for a product.

## Changes
- Removed possibility to leave a star rating in Shop page and added it to be inside detailed view only
- Added Review field in backend
- Added tests to work with new review field
- Added a comment-like section for other people's reviews
- Added ability for user to leave a text review on-top of a simple star rating.
**Before**:
Shop View:
image
Detailed View:
image

**After**:
Shop View:
image

Detailed View:
image


J91-cloud (approved):

Valthefirst (approved):
LGTM the ratings feature still works and I can leave a review. I was able to switch users and still see the review left by the previous account. Also, I know it's totally unrelated but the nav bar now has 2 Shops buttons
image

PR #902, opened 2024-10-07
bug(VIST CPC 1301) Fixed the problem and added an obj folder to the gitignore

FelixAllard:
**JIRA:** link to jira ticket
[## Context:](https://champlainsaintlambert.atlassian.net/browse/CPC-1301)

I am doing those changes so that the emailing tab works without the need to send a first email
No
## Changes
Changed the request in the front end to handle NoContent
Added emailing-service-test/ obj folder to gitignore
No UI


Rrregine (approved):
I see you remove the Obj folder from the testing which is good practice. 

29 files are changed but actually only 1 was changed.

ioannisp03 (approved):
Saw that you added the obj of testing to the gitignore.

PR #911, opened 2024-10-08
bug(PROD-CPC-1302): productQuantity is Null

Valthefirst:
**JIRA:** [link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-1302)
## Context:
When adding a product the productQuantity would always be set to null, so we needed to fix that bug.
Where's Jared?

## Changes
I just added a line of code to map the productQuantity from the productRequestModel to the Product entity
There are no UI changes

### DIFF
--- a/products-service/src/main/java/com/petclinic/products/utils/EntityModelUtil.java
+++ b/products-service/src/main/java/com/petclinic/products/utils/EntityModelUtil.java
@@ -33,6 +33,7 @@ public class EntityModelUtil {
                 .productSalePrice(productRequestModel.getProductSalePrice())
                 .averageRating(productRequestModel.getAverageRating())
                 .productType(productRequestModel.getProductType())
+                .productQuantity(productRequestModel.getProductQuantity())
                 .build();
     }
 


viktorkuts (approved):
This fixes the null productQuantity. 
However now there is a new bug (not caused by this PR), where the product in Recently Clicked Products is not updated when the quantity is updated, I'll make a new issue for it on Jira.
(Look how the fish tank heater has a good product quantity as the recently one has a out of stock)
image

Valthefirst:
> This fixes the null productQuantity. However now there is a new bug (not caused by this PR), where the product in Recently Clicked Products is not updated when the quantity is updated, I'll make a new issue for it on Jira. (Look how the fish tank heater has a good product quantity as the recently one has a out of stock) image

Probaly cause the recently clicked products doesn't always fetch the product when refreshing the page

J91-cloud:
> This fixes the null productQuantity. However now there is a new bug (not caused by this PR), where the product in Recently Clicked Products is not updated when the quantity is updated, I'll make a new issue for it on Jira. (Look how the fish tank heater has a good product quantity as the recently one has a out of stock) image

This is because when applying the the recently clicked feature the developer only keeps the state of the previous item when clicked. It will not change the card css if the state is changed without clicking it.

J91-cloud (approved):

PR #921, opened 2024-10-09
bug(VETS-CPC-1390): update delete vet profile photo button

bluement:
**JIRA:** link to jira ticket: https://champlainsaintlambert.atlassian.net/browse/CPC-1390 
## Context:
before, when the profile picture was default you still had the "delete" button, which was bad for ux because you couldn't actually delete the default picture
## Does this PR change the .vscode folder in petclinic-frontend?: no

## Changes
only changed one file: vet details.tsx
before:
image
After:
image

## Dev notes (Optional)
i changed some lines of the update method, but it works perfectly
Pull request links


hristoiv11 (approved):
The changes that have been made, contribute to the logic of the project. Good job.

orell2j (approved):
Thank you for making such a change, now it looks aesthetically better, I hope the button works and isn't visible after the first delete.

PR #924, opened 2024-10-10
feat(PROD-CPC-1366): product details

viktorkuts:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-1366
## Context:
The product details page looked very bland. This PR attempts to fix some problems with product details and modernize the product details page.
## Changes
- StarRating component changed to only use numbers instead of RatingModel
- StarRating component changed to accommodate floating point values (partial stars)
- ProductDetails now follows a REST format inside the URL of the document page
- Added character limit checking in ReviewBox
- Updated Product card to use StarRating
- Made navbar only have one shop link
- Some refactoring
- Remade whole layout of the product details page
- Added styling to product details page
- Edit and delete buttons only show for admins
**Before**:
image
image

**After**:
image
image
## Dev notes (Optional)
Delete and add to cart have not been implemented as it is out of my story's scope, pops an alert saying not yet implemented.
Note that the StarRating component now has a new property *viewOnly*, which dictates whether it is a component that is going to be used for clicking or only for rendering stars. The updateRating property is now optional, only if viewOnly is set to true.


petclinic-frontend/src/shared/models/path.routes.ts
```diff
@@ -48,8 +48,8 @@ export enum AppRoutePaths {
   SignUp = '/users/signup',
   CustomerVisits = '/customer/visits',
   UpdatePet = '/pets/:petId/edit',
-  ProductDetails = '/productDetails',
-  EditProduct = '/products/edit',
```
J91-cloud, line 52: Thank you for adding this. I will sleep peacefully tonight.

J91-cloud (approved):
Great Job Victor! Personally i think the stars a little big when in the main page of the products but a part from that it is night and day difference then what we had before.

ZacharyLelievre (approved):
Great job updating the product details page! The new layout and styling look much better, and the changes to the StarRating component and the character limit for reviews are great additions. Maybe consider adding an alert when the limit is reached, but other than that, great job!

PR #928, opened 2024-10-12
bug/(PROD-CPC-1399): Updating Quantity Issue

ZacharyLelievre:
**JIRA:** [link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-1399)
## Context:
The Jira ticket fixed an issue where updating the product quantity gave wrong results. When users tried to increase the quantity by one, it doubled (from 4 to 9 instead of 5). Also, reducing the quantity by any amount would  set it to 1 automatically.
No, this PR does not make any changes to the .vscode folder.

## Changes

- Fixed the issue with product quantity being doubled during updates.
- Fixed the problem with product quantity being set to one when reducing by any number during updates of the quantity.

### Before when updating quantity:
image
### Adding one would result to this:
image
### Removing one or any number would result to the value of 1:
image

### Now Fixed:
image
### Now when adding one:
image
### Now when removing one:
image
## Dev notes (Optional)
Pull request links


viktorkuts (approved):
Looks good, quantity updates with expected behavior (sets the absolute value that is given to it instead of doing an addition/subtraction).

ThomasBedard (approved):
Works great, simple code fix nothing much to add. 👍

PR #930, opened 2024-10-12
 bug/(PROD-CPC-1409):Fixed-FrontEnd_CSS

J91-cloud:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-1409

## Context:

VETS merged something that effected our service , i made the changes so now both work well.

Before: It was in a flex direction and nothing was working properly.

Screenshot 2024-10-12 112407


Edwxd (approved):
VetListCard files have the necessary changes. I should have specified the class name for the vet card as "card-vets" instead of "card" alone.  A mistake on my part for bad naming conventions. The code should run fine. Thanks for taking care of it.

viktorkuts (approved):
Looks good, vets and shop pages look like what they are intended to look like. Other pages don't seem to be affected by the change. Thanks for taking care of this bug.

PR #934, opened 2024-10-13
feat(CART-CPC-1338): move products between the cart and wish list

Hares-2088:
**JIRA:** [link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-1338)
## Context:
Now, a user can move his products between his cart and his wishlist. When it moves from one to the other, it is removed from the latter. 
No it does not

## Changes
- The react app fetches correctly the cart service for these two methods: moveToCart and MoveToWishList
- Each button is only displayed when it is in the right section. this means that for example, there is not going to be a button move to wishlist when the product is already in the wishlist
- the useEffect is used to fetch a get all products each time the user is going to move a product in his cart


mzolovska (approved):
It looks good, from what I understand is that you improved the wishlist function, and it is much more realistic now!

Zako563 (approved):
The code is good and I dont see any mistake. I saw that you modified the front end but you dont share any screenshot in your pr. I think you need to do it for grade purpose. Other than that good job.

PR #939, opened 2024-10-13
bug/(PROD-CPC-1386): Rating Filter No Error Message

ZacharyLelievre:
**JIRA:** [link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-1386)
## Context:
The ticket is about an issue where no error message was shown when a user would set the minimum rating higher than the maximum rating on the product page. This was confusing, so I added an error message to fix it.
No changes to the .vscode folder in this PR.
## Changes

- Added a rating validation to display an error message when the minimum rating is higher than the maximum rating.
- Added a disabled function to prevent a user to click apply when setting the min star higher than the max.

Before:
image
After:
image


Valthefirst (dismissed):
It's nice that there's a message telling them that's not permitted 
(nitpic) just a design suggestion maybe let the users set a minium star and let them leave the maxium empty

viktorkuts:
Looks good, I can see the error message, however it doesn't prevent me from clicking on Apply. If you're able to also disable the apply button if the error is present, that'd be a great addition.

ZacharyLelievre:
> Looks good, I can see the error message, however it doesn't prevent me from clicking on Apply. If you're able to also disable the apply button if the error is present, that'd be a great addition.

Good idea! it has been implemented now using the "disabled" function to disable the button

ThomasBedard (approved):
Works perfectly 👍 

Screenshot 2024-10-17 at 11 07 48 PM

Elvis-elvis (approved):
Well implemented Zachary

PR #942, opened 2024-10-15
bug/breaking_change_with_gradle

nic5694:
**JIRA:** link to jira ticket
## Context:

## Changes
## Dev notes (Optional)
Pull request links


viktorkuts:
While you're at it, do you think you can also modify the vet-service check?
It looks like vet-service doesn't have a working directory assigned to it, so it runs all of the services tests. (I had an error in product service but vet service was failing because it tried running product-service tests)
This would explain why vets check is taking a long time when running checks.
Refer to [this](https://github.com/cgerard321/champlain_petclinic/actions/runs/11371648256/job/31634186743?pr=944), it's not supposed to run product-service.

nic5694:
> While you're at it, do you think you can also modify the vet-service check? It looks like vet-service doesn't have a working directory assigned to it, so it runs all of the services tests. (I had an error in product service but vet service was failing because it tried running product-service tests) This would explain why vets check is taking a long time when running checks. Refer to [this](https://github.com/cgerard321/champlain_petclinic/actions/runs/11371648256/job/31634186743?pr=944), it's not supposed to run product-service.

Yes It is a problem with the vet service, it's because the gradle was not set up proprely and intially it didn't have a gradlew file. Their is already a ticket, you can pick up the ticket if you would like. It just involves reinitializing gradle. You can do it with intellij also or within the command line if you install the gradle cli.

PR #944, opened 2024-10-16
bug(PROD-CPC-1411): average rating unexpected behavior fix

viktorkuts:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-1411
## Context:
Average rating seemed to be incorrectly implemented inside the DataLoader. This PR fixes this by adding Ratings in the DataLoader.
## Changes
- Added Ratings to DataLoader
- Removed hardcoded averageRatings in the Product entity
- Reverted updateProduct method to commented out one
## Dev notes (Optional)
As the updateProduct has been reverted, please verify all edit methods work on your end. I have tested them and they work, but please do the due diligence to test if it didn't break anything.

PR #953, opened 2024-10-17
feat(PROD-CPC-1348): Dropdown Menu List 

Elvis-elvis:
**JIRA:** link to jira ticket
https://champlainsaintlambert.atlassian.net/browse/CPC-1348
## Context:
This ticket adds a dropdown menu to the filter section, enabling guest users to easily filter products by type. The goal is to improve user experience by offering a convenient way to browse products, especially for those unfamiliar with the available types.

## Changes
- Added a dropdown menu in the filter section of the frontend to allow users to select and filter products by type.
- Integrated backend support to fetch products filtered by ProductType through a new API endpoint (/api/v1/products/filter).
- Updated ProductController with a new method to handle product filtering by ProductType.
- Implemented getProductsByType(ProductType productType) to retrieve products based on the selected type.
- Updated the frontend to display product types dynamically in the dropdown menu and handle filtering requests when a user selects a type.
image
image
image

## Dev notes (Optional)
Pull request links


ZacharyLelievre (dismissed):
The feature works as intended, but there are issues when trying to filter by product type along with other filter options. This is because the filtering was implemented outside of the GetAll logic. However, a PR bug fix is expected to address this part.

products-service/src/main/java/com/petclinic/products/presentationlayer/products/ProductController.java
```diff
@@ -85,5 +89,13 @@ public Mono<ResponseEntity<Void>> changeProductQuantity(@PathVariable String pro
                 .flatMap(request -> productService.changeProductQuantity(productId, request.getProductQuantity()))
                 .then(Mono.just(ResponseEntity.noContent().build()));
     }
+    @GetMapping("/filter")
+    public List<ProductResponseModel> getProductsByType(@RequestParam ProductType productType) {
+        List<Product> products = productService.getProductsByType(productType);
+        // Convert the entities to response models (if needed)
+        return products.stream()
```
viktorkuts, line 96: Instead of using products.stream, it's best to use a Reactive Flux in Repository and Service.

products-service/src/main/java/com/petclinic/products/presentationlayer/products/ProductRequestModel.java
```diff
@@ -15,6 +16,7 @@ public class ProductRequestModel {
     private String productName;
     private String productDescription;
     private Double productSalePrice;
-    private String productType;
+    private Double averageRating;
```
viktorkuts, line 19: I removed averageRating from the ProductRequestModel in [my PR](https://github.com/cgerard321/champlain_petclinic/pull/944) to fix any averageRating overrides. If you could remove the averageRating field from requestModel please.

products-service/src/main/java/com/petclinic/products/utils/DataLoaderService.java
```diff
@@ -40,7 +41,8 @@ public void run(String... args) throws Exception {
                 .productDescription("Premium dry food for adult dogs")
                 .productSalePrice(45.99)
                 .requestCount(0)
-                .productType("Food")
+                .averageRating(1.3)
```
viktorkuts, line 44: Same goes here, I fixed these issues in [this PR](https://github.com/cgerard321/champlain_petclinic/pull/944) and this would essentially bring back the Rating bug.

products-service/src/main/java/com/petclinic/products/datalayer/products/ProductRepository.java
```diff
@@ -22,6 +24,7 @@ public interface ProductRepository extends ReactiveMongoRepository<Product, Stri
     @Query("{ 'productSalePrice' : { $lte: ?0 } }")
     Flux<Product> findByProductSalePriceLessThanEqual(Double maxPrice);
     Flux<Product> findProductsByProductType(String productType);
+    List<Product> findByProductType(ProductType productType);
```
viktorkuts, line 27: Convert the List to a Flux<Product>, that way you can use Reactive streams instead of manually creating them in the controller

viktorkuts (changes requested):
Feature still needs some work:
1. Edit page does not have a dropdown
image
# Bugs:
1. Please remove instances of averageRating in the RequestModel and in Tests, since this was fixed in [previous PR](https://github.com/cgerard321/champlain_petclinic/pull/944)
2. Filtering results in some fields of the Product not returning, causing issues:
image

About Reactive Streams, it's more of a nitpick than anything, it works fine, but it is so the implementations look the same.

petclinic-frontend/src/features/products/ProductsList.tsx
Valthefirst, line 1: Why are you hard coding the values? What if the manager wants to add new types? Will they have to learn how to code and edit it? It's kinda user unfriendly. You could fetch the types from a database that stores them like that when someone adds a new one there's no need to change code in the front end

Valthefirst (changes requested):

petclinic-frontend/src/features/products/ProductsList.tsx
Elvis-elvis: Thanks for the review Valentine. I changed the label which was hard coded
image

products-service/src/main/java/com/petclinic/products/presentationlayer/products/ProductRequestModel.java
Elvis-elvis: Thanks for the review Victor. I think my main wasn't updated when I was coding but I fixed and didn't have the fixed bug you did but I fixed and removed the averageRating

petclinic-frontend/src/features/products/ProductsList.tsx
Valthefirst: That's great👍🏾but can you answer this question why are you using enums instead of making a new database to store the enums or a List? You cannot dynamically add new values to an enum in Java at runtime. You definetly don't have all of the possible product types. You need to let the manager the possiblity to add new ones.

Valthefirst: But please use the database option, it's much more simpler

Elvis-elvis: Thanks for the follow-up! You're right, enums are static and don't allow dynamic updates, which limits flexibility for adding new product types. While a dynamic database solution would provide more flexibility for adding product types in the future, it's not a current requirement, and enums offer simplicity for now without adding complexity

ThomasBedard (approved):
As mentioned previously, it doesn't utilize the database but rather enums, however this works for now. We might need to add medications as right now only the flea collar is in that type and isn't a medication.

Screenshot 2024-10-17 at 11 53 59 PM

viktorkuts (approved):
Changes about the previous rating bug pr have been made. If the bug with the null fields will be fixed in the next bug pr, then it's ok I'll approve.

J91-cloud (changes requested):
image

## Problem

Adding a product no longer works. I think this is because i can put any type of product. In the example above i put sssss. Make sure that the person who is adding a product can only put a specific type that is available and exists.

J91-cloud (approved):
Obvious bugs but the developer will solve them later in future PR's.

Valthefirst (approved):
The dropdown list works as intended and since @J91-cloud said the bugs will be handled in a future PR I will approve this. But just know @viktorkuts @ThomasBedard @ZacharyLelievre @J91-cloud @Elvis-elvis if this passes it will go against our Sprint goal

ThomasBedard (approved):
As mentioned before, there are some issues with this PR, it will be fixed in the future.

Valthefirst:
> As mentioned before, there are some issues with this PR, it will be fixed in the future.

It was just warning I don't really care at this point

PR #956, opened 2024-10-17
Updated steps to run if failed

viktorkuts:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-1413
## Context
When Qodana fails, the build report is not packed and archived. This PR attempts to run the archive of build reports even if Qodana build fails.

## Changes
- Run archive step when Qodana succeeds or fails

### DIFF
--- a/.github/workflows/qodana_code_quality.yml
+++ b/.github/workflows/qodana_code_quality.yml
@@ -31,11 +31,13 @@ jobs:
         run: ./gradlew build
 
       - name: Create coverage-data directory and copy contents
+        if: success() || failure()
         run: |
           mkdir -p .qodana/code-coverage
           cp -R */build/reports/* .qodana/code-coverage/
 
       - name: Archive coverage data
+        if: success() || failure()
         uses: actions/upload-artifact@v4
         with:
           name: gradle-coverage-data.zip


viktorkuts:
Artifact is created even if qodana fails
image

nic5694:
i would suggest changing it to a feature it was not a bug. It is normal that the task would not run if the build task has failed. It would only be a bug if you had a condition and that condition was true and it was not running

nic5694:
@viktorkuts Here follow this, i think uploading the HTML file might cause issues, this action will do everything you want: https://github.com/marketplace/actions/jacoco-report

viktorkuts:
> @viktorkuts Here follow this, i think uploading the HTML file might cause issues, this action will do everything you want: https://github.com/marketplace/actions/jacoco-report

Yes I've seen this action, however it uses XML and I wanted to try parsing HTML reports beforehand (didn't really work). I'll enable XML reports, seems to be simpler after experimenting, thanks.

viktorkuts:
There is a bug within jacoco-report action, which shows services not meeting the min-coverage as passing. 1.4 should have fixed this, however the feature where we need to use multiple reports is present in the latest release, 1.4 errors.
Here is the [issue](https://github.com/Madrapps/jacoco-report/issues/86)

PR #913, opened 2024-10-08
feat(INVT-CPC-1385) Delete all supplies from an inventory

igouriny:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-1385
## Context:
What is the ticket about and why are we doing this change.
Implemented the ability for a vet to delete all of the supplies inside of a specific inventory
## Does this PR change the .vscode folder in petclinic-frontend?:
No, it does not
## Changes
- Added controller methods in both the service and the API-Gateway
- Added the methods in the ProductInventory service and the ProductInventoryServiceImpl
- Added tests inside of the service and the api-gateway (100% coverage for both)
- Added a button to the front-end
- Added an api method that uses axios to use the back-end method.
## Before and After UI (Required for UI-impacting PRs)
Before:
Screenshot 2024-10-08 at 7 13 45 PM
After:
Screenshot 2024-10-08 at 7 14 08 PM
Screenshot 2024-10-08 at 7 14 27 PM
Screenshot 2024-10-08 at 7 14 40 PM

PR #977, opened 2024-10-19
bug(CPC-INVT-1416): Fix supply inventory page

igouriny:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-1416
## Context:
There is a bug where if the supplies inside an inventory are set to 0 (empty), the inventory wouldn't render and not allow the user to do necessary requests.
No, it does not.

## Changes
- Removed the error message from the fetching inventories API call
- Added a message that says there are no supplies inside the current inventory.
Before:
Screenshot 2024-10-18 at 10 24 05 PM

After:
Screenshot 2024-10-18 at 9 57 19 PM

https://github.com/cgerard321/champlain_petclinic/pull/913 
& any other PR that used to be affected by not being able to access inventories when there were no supplies

### DIFF
--- a/petclinic-frontend/src/features/inventories/InventoryProducts.tsx
+++ b/petclinic-frontend/src/features/inventories/InventoryProducts.tsx
@@ -68,8 +68,6 @@ const InventoryProducts: React.FC = () => {
         setProducts(response.data);
         setProductList(response.data); // Set productList as well
         setFilteredProducts(response.data); // Initialize filtered products with all products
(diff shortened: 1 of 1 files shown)


Carlos-123321 (approved):
Made it so when the user clicks on an inventory card which contains zero products it shows a page where the user can add products to the inventory clicked. Great addition.

Sunveerg (approved):
Quick and clean fix good job!

PR #987, opened 2024-10-19
bug(INVT-CPC-1417): make uploaded images not required

Carlos-123321:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-1417
## Context:
Everytime we want to update our inventory we must always select an image because our uploaded image input box requires a value and it doesn't already have a value. This is because logic isn't implemented to fetch the uploaded image from the database and store that value fetched into the input box. But this isn't needed thanks to this simple fix.
**No, it doesn't**
## Changes
• Removed required from both the add inventory form and edit inventory form
## Dev notes (Optional)
Pull request links

### DIFF
--- a/petclinic-frontend/src/features/inventories/AddInventoryForm.tsx
+++ b/petclinic-frontend/src/features/inventories/AddInventoryForm.tsx
@@ -178,7 +178,6 @@ const AddInventoryForm: React.FC<AddInventoryProps> = ({
               id="imageUpload"
               accept="image/*"
               onChange={handleImageUpload}
-              required
             />
           </div>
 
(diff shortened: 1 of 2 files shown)


petclinic-frontend/src/features/inventories/AddInventoryForm.tsx
```diff
@@ -178,7 +178,6 @@ const AddInventoryForm: React.FC<AddInventoryProps> = ({
               id="imageUpload"
               accept="image/*"
               onChange={handleImageUpload}
-              required
```
CharlesS42, line 181: Taking out the 'required' keyword fixes the image bug noticed after the upload image feature implementation. Well spotted.

CharlesS42 (approved):

Th0masL3 (approved):
Simple fix that enhances the UX. Simple yet needed. Good fix!

PR #989, opened 2024-10-19
feat/(INVT-CPC-1344): improve mainpage user interface and experience

Carlos-123321:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-1344
## Context:
We are removing the buttons we had on the bottom of our inventory page and instead putting those buttons inside a menu which the users can click to see actions they can perform.
**No, it doesn't**
## Changes
•Removed the buttons located in the bottom of our page
•Created a menu icon which can be clicked by the user to see actions that can be performed

**Before** this is how the bottom of our page looked like:

image

**After** this is how the bottom of our page looks like:

image

**After** this is the menu which users can click to see actions they can perform:

image

## Dev notes (Optional)
Pull request links

### DIFF
--- a/petclinic-frontend/src/features/inventories/InventoriesListTable.module.css
+++ b/petclinic-frontend/src/features/inventories/InventoriesListTable.module.css
@@ -1,3 +1,42 @@
+#menuSection {
+  display: flex;
+  flex-direction: row;
+  margin-bottom: 50px;
+}
+
+#menuContainer {
+  width: fit-content;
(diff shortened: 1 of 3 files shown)


igouriny (approved):
Looks good! I like the use of the hamburger menu.

Sunveerg (approved):
The color of the buttons might still have to be changed but it definitely looks cleaner now good job!

PR #992, opened 2024-10-20
bug(VETS-CPC-1415): setup vet service context test and update versions

kyrusho:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-1415

## Context:
This ticket appears to be about setting up a basic integration test for the VetServiceApplication in a Spring Boot application. The goal is likely to ensure that the application is correctly configured for integration testing and that the context can be loaded in a test environment. It also updates Gradle and other versions to be up-to-date with other services.

No, it does not change the .vscode folder.

## Changes
Added Annotations:
- @SpringBootTest: This annotation indicates that the application context should be loaded for integration tests.
- @ActiveProfiles("test"): This specifies that the test profile should be active during the tests, allowing for different configurations.

Added a Test Method:
- void contextLoads(): This is a simple test method to check if the application context loads successfully.

Class Declaration:
- The class is now defined with annotations above it, which is typical for a Spring Boot test class.

- Updated Gradle version from 7.3.3 to 8.3.

No UI changes have been made.


hristoiv11 (approved):
The integration test setup looks good. The application context should load successfully with the test profile. No issues are seen during the review.

Apoleh (approved):
Thanks for updating our versions and making the context load test, it's nice you noticed. Rest looks good to me.

PR #998, opened 2024-10-20
Feat/prod cpc 1412 delete product2

J91-cloud:
**JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-1412

## Context:
Added the delete function capability, now when we delete an object we return to productList page. 

## Changes

##Before

>[!Note]
When deleting we had a localhost error saying the feature has not been implemented yet

## After 
image


Valthefirst (approved):
The feature works well as intended and the deleted product is gone from the database.
(nitpic) I know it wasn't part of your requirement but it would also be nice if an inventory manager would be able to delete too

ZacharyLelievre (approved):
The feature works as expected, and the product deleted is functioning correctly. No issues found. Good work Jessy!

PR #1023, opened 2024-11-09
config/e2e playwright initial setup

cgerard321:
**JIRA:** 
none
## Context:
Added basic configuration for playwright to be used as an e2e testing tool. It is not yet connected to the CI pipeline.
Yes. I added the playwright extension to extensions.json file.

## Changes
- updated tsconfig.json to work with playwright tests
- added playwright test for vscode
- updated default playwright.config.ts 
- added various test files containing example test cases for petclinic

n/a 
## Dev notes (Optional)
This is a very basic installation. In the future, it needs to be updated with the following:
- automatic login of admin and other roles
- preset locators for all the different pages
- anything to make the tests simpler and more fool-proof to create

none

### DIFF
--- a/petclinic-frontend/.gitignore
+++ b/petclinic-frontend/.gitignore
@@ -25,3 +25,7 @@ dist-ssr
 !/.vscode/settings.json
 !/.vscode/extensions.json
 .vscode/
+/test-results/
+/playwright-report/
+/blob-report/
+/playwright/.cache/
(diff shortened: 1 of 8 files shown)


nic5694 (approved):
Everything works, except for the add user test i would suggest running it with the option show in browser so you can click the customers button. Their is issues with the clicking and the state i think the way that it is set up in the frontend is problematic a ticket should be opened for that.

PR #1026, opened 2024-11-11
E2etest/cust justin morissette

Justin222993:
# Testing for class
Added test for getting user by ID and seeing if the correct information is being displayed


FelixAllard (approved):
Yeah, seems good to me. A single file modified, looks like good work

flyindonut (approved):
Amazing test, it is so in depth. Great job Justin!

PR #1038, opened 2024-11-11
e2etest/PROD-Thomas

ThomasBedard:
## Context:
Made an e2e test for edit a product name
it doesn't


Valthefirst (approved):
Your test works fine the only wickup is that is can't be run twice without re composing up

ZacharyLelievre (approved):
Well implemented! Good Job!

PR #1044, opened 2024-11-11
test for get by name

Apoleh:
test to search vet by name


cgerard321:
Remove incorrect folders and files from this PR. There should only be 1 file changed with is the file containing your test.

PR #1048, opened 2024-11-11
e2etest/invt-ilyassg: Added E2E test for deleting supplies in inventories

Edwxd:
Added E2E test for deleting supplies in inventories (Ilyass)


J91-cloud (approved):
Great Test

PR #1049, opened 2024-11-11
e2etest/VETS-edwardnasser: Created a end to end test for updating vet information

Edwxd:
## Context:
Created a end to end test for updating vet information. 

## Does this PR change the .vscode folder in petclinic-frontend?
 No it does not.

## Changes

- Added the test file for update vet information


J91-cloud (approved):
Great work

PR #1050, opened 2024-11-11
Product Search Bar

J91-cloud:
**JIRA:** link to jira ticket
## Context:

`Search method. When typing h in the search bar, horse saddle must be showed. `


Edwxd (approved):
Test looks good. Great Job!

PR #1051, opened 2024-11-11
Test for updating an inventory

Carlos-123321:
**JIRA:** link to jira ticket
## Context:
Added a test which checks that if an inventory is updated correctly, it should display a message saying "Inventory updated successfully".


vinivelozo (approved):

Sunveerg (approved):
Update test looks good

PR #1060, opened 2025-09-05
bug/Added checks to db init

DylanBrass:
**JIRA:** N/A
## Context:
When deploying the services to PROD, I noticed an issue where duplicate keys would be added.
No Nic, it does not :)

## Changes
- Added checks to the repos of each service if elements are present, to not insert in the db.

This is not perfect, but will be more than enough for school.

### DIFF
--- a/petclinic-frontend/src/features/inventories/api/getAllInventoryTypes.ts
+++ b/petclinic-frontend/src/features/inventories/api/getAllInventoryTypes.ts
@@ -3,7 +3,7 @@ import { InventoryType } from '@/features/inventories/models/InventoryType.ts';
 
 export async function getAllInventoryTypes(): Promise<InventoryType[]> {
   const response = await axiosInstance.get<InventoryType[]>(
-    axiosInstance.defaults.baseURL + 'inventories/types'
+    axiosInstance.defaults.baseURL + 'v2/gateway/inventories/types'
   );
   return response.data;
 }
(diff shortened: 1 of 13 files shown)


cgerard321 (approved):

cgerard321 (dismissed):

cgerard321 (approved):

petclinic-frontend/src/features/inventories/api/getAllInventoryTypes.ts
```diff
@@ -3,7 +3,7 @@ import { InventoryType } from '@/features/inventories/models/InventoryType.ts';
 
 export async function getAllInventoryTypes(): Promise<InventoryType[]> {
   const response = await axiosInstance.get<InventoryType[]>(
-    axiosInstance.defaults.baseURL + 'inventories/types'
+    axiosInstance.defaults.baseURL + 'v2/gateway/inventories/types'
```
nic5694, line 6: does this work? their is supposed to be a v2 flag passed to the shared axios instance

petclinic-frontend/src/features/inventories/api/searchInventories.ts
```diff
@@ -20,7 +20,7 @@ export async function searchInventories(
     : `inventories?page=${currentPage}&size=${listSize}`;
 
   const response = await axiosInstance.get<Inventory[]>(
-    axiosInstance.defaults.baseURL + url
+    axiosInstance.defaults.baseURL + 'v2/gateway/' + url
```
nic5694, line 23: same thing here, i just checked look at the sharedAxios instance file in the shared/api folder

nic5694 (approved):

petclinic-frontend/src/features/inventories/api/searchInventories.ts
DylanBrass: Same as above, will be fixed by other tickets

PR #1065, opened 2025-09-18
fixed put inventory endpoint for the customer frontend

christianJames24:
it was missing a / because without it, it would result in "/api/v2/gatewayinventories/" instead of "/api/v2/gateway/inventories/"

**JIRA:** link to jira ticket
## Context:

## Changes
## Does this use the v2 API?:
## Dev notes (Optional)
Pull request links


megadoxs (changes requested):
Your PR name doesn't follow the convention.
You should properly fill the PR template.
The api version should be specified in the arguments to the shared axios instance.

PR #1068, opened 2025-09-19
bug/(BILL-CPC-1440): Fix_Improper_Use_Of_API_Calls_In_FE

VladLoghin:
…of hardcoded urls.

**NOTE
- The post request on the admin page to create a new bill creates a new bill with the first name being null, I will create a ticket on jira to fix this issue
- The get by billId is not working, I will create a ticket on jira to fix this issue
- I changed the getBillsByCustomerId method in the CustomerBillController to return APPLICATION_JSON_VALUE, since I couldn't get the front-end to handle SSE with axios on this one. I will create a ticket on jira to fix this issue

**JIRA:** link to jira ticket
## Context:
The API's being hardcoded causes the hosted front-end to not work.

## Changes
- The front-end API calls now use the axios format provided by Dylan.
## Does this use the v2 API?:
Yes because some of the requests in the front-end were using v2
NO
No changes
## Dev notes (Optional)
The getBillsByCustomerId in CustomerBillController was changed from text-event stream to application json because I tried to make it handle SSE but failed and would rather get my changes up on main as soon as possible to get on to my next task


megadoxs (changes requested):
Please follow the naming convention for PRs.
withCredentiats doesn't need to be specified since the shared axiosInstance has it set to true by default.
An api version should always be specified when using the axiosInstance to follow the contributing guidelines.

megadoxs:
Good jobs, although you have not renamed the pr to follow the naming convention. You should also look into the changes proposed by copilot. From what i can tell without diving to deeply in the code at the very least the last one about a flux<flux> looks resonable. For the others I would need to look at the code in more details, but you should still investigate them.

TMBJake:
Yeah it seems copilot wants a couple of changes, check them out in more detail, I looked at them quickly and some look like reasonable changes. Resolve those please.

TMBJake (approved):
This ticket is DONE:
followed proper code,
Fixed main issue at hand and added tickets for any found bugs,
handled all changes properly,
Proper naming conventions,
no issues with tests.

DeanteMirijello (approved):
Changes look great!

DeanteMirijello (approved):
Looks good

PR #1071, opened 2025-09-20
bug(INVT-CPC-1437): Fix Inventory Api Calls 

OlivierGingras026:
**JIRA**: https://champlainsaintlambert.atlassian.net/browse/CPC-1437

**Context:**

This PR focuses on refactoring and stabilizing the inventory module across the API Gateway and frontend. The goal is to centralize all inventory-related logic into dedicated controllers, ensure frontend calls are consistent with backend endpoints, and improve maintainability by consolidating tests.

No

Reviewers need to check for any changes to the
.vscode folder and add a comment about it to their review comments.

**Changes**

**Refactored inventory endpoints**

- Moved all inventory-related endpoints into a new controller in a new V1 folder so they can be used by everyone.

- Added the breaking v2 endpoints into the same controller to ensure backward compatibility and avoid breaking changes.

**Frontend fixes**

- Removed hardcoded values in the frontend inventory implementation.

- Updated inventory API calls to use the shared axiosInstance for consistent configuration and request handling.

**Test refactoring**

- Consolidated all inventory-related tests into the InventoryControllerV2Test class.
-
- Updated the test structure to align with the new controllers and API endpoints.

Does this use the v2 API?:
No, as mentioned, I removed all v2 API from v2 inventory controller and put them in v1 if necessary(avoid issue with breaking changes)

No


megadoxs:
I don't think you were meant to remove api calls from the BFFApiGatewayController and move them to a InventoryControllerV1 in a new directory.

megadoxs:
I would have to ask them again, but that is not how I understood it.

megadoxs (dismissed):
Try catchs are redundant when using axiosInstance since it already catches its own exeptions.

OlivierGingras026:
She told me this "Furthermore, this is a good time to start a separate v1 controller for your team's endpoints like Dylan showed in his talk."

OlivierGingras026:
> Try catchs are redundant when using axiosInstance since it already catches its own exeptions.

You're kinda right, but in the Front End checklist, it was mentioned this "Includes error handling with try-catch"

megadoxs:
I have asked dylan about the api and nic about the try-catchs to clarify things. if I am wrong I will remove my change request.

DylanBrass:
The way things were separated is exactly what I said in class. I didn't review the code a lor, but the endpoints look the same to me. Be sure to test the new and old frontend to verify that nothing broke. 

For the try catch, it should be done in axiosInstance, where it redirects to a page if needed. But errors like 404 shouldn't redirext you if you fetch unknown type, for example.

nic5694:
> The way things were separated is exactly what I said in class. I didn't review the code a lor, but the endpoints look the same to me. Be sure to test the new and old frontend to verify that nothing broke.
> 
> For the try catch, it should be done in axiosInstance, where it redirects to a page if needed. But errors like 404 shouldn't redirext you if you fetch unknown type, for example.

The way the instance was initially set up it would catch every single error, that configuration got written over and doesn't work exactly. But analyzing the requirements I think that using a hybrid approach would be best. 

As the global error handling stops propagation so it will catch errors like 401, 403, 500, 502, 503, 504 and redirect users or show system wide error pages. 

The Error handling continues to components as status codes like 
```
// These should be handled by individual components
400 → Bad request (show form validation)
404 → Not found (show "item not found")
409 → Conflict (show specific conflict message)
422 → Validation errors (show field errors)
429 → Rate limiting (show "slow down" message)
```

To use this in your components it would be as follows:
Global errors will be automatically handled:
```typescript
// No try-catch needed - 401/500 etc. are handled globally
const loadUserProfile = async () => {
  const response = await axiosInstance.get('/user/profile');
  setUser(response.data);
};
```
Local errors will require a try-catch:
```typescript
// Handle validation errors locally
const updateUser = async (userData) => {
  try {
    const response = await axiosInstance.put('/user', userData);
    setUser(response.data);
    showSuccess('User updated!');
  } catch (error) {
    // 400/422 validation errors come here
    if (error.response?.status === 422) {
      setFieldErrors(error.response.data.errors);
    } else if (error.response?.status === 404) {
      showError('User not found');
    }
  }
};
```

Overall critical errors will be handled consistently throughout the application. Business logic errors will be dealed with where their context exist and their will be a clear separation between global and local concerns.
@megadoxs I sent you a discord message with the updated axios instance and axios error handler to test this

OlivierGingras026:
So, do I keep it as it is right now, and next sprint, I create a ticket for the try/catch exception ?

PR #1089, opened 2025-09-22
bug(PROD-CPC-1335): fixed filter feature to work properly

Dnb182018:
**JIRA:** 
[link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-1335?atlOrigin=eyJpIjoiMDNiMzZlZDQ3N2I5NDllYWIzMGFmOWIyMjM1YTIwN2IiLCJwIjoiaiJ9)
## Context:
Products list and its details should be list out according to selected filter. Currently, the list display incorrectly when an item types is selected.
No

## Changes

- Items type was using getProductsByType() but it didn't work with other filter condition so i added a productType in the paramater of the getAllProducts().
- ProductList.tsx , ProductModel.ts and getAllProducts.ts was changed because Items types box recently running seperately with getProductsByType while all other filter feature is in getAllProducts.

## Does this use the v2 API?:
Before:
SortFeatureBefore
After:
SortFeatureAfter
## Dev notes (Optional)
Pull request links


Ayoub-Seddik (approved):
Approved, since you demonstrated that all the bugs regarding the feature and sort by are fixed

PR #1093, opened 2025-09-23
feat(CART-CPC-1457): auto-refresh cart notification dot after add/remove

LePingouins:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-1457

## Context:
The cart notification badge in the top NavBar does not update automatically after adding or removing items; users must refresh the page. This PR makes the badge react to cart changes so it always reflects the current item count.

No changes to .vscode.

## Changes
- Frontend (petclinic-frontend):
     - Added a new effect that recalculates the cart item count every time the cart items change.
     - Removed the old manual call that tried to update the cart count right after fetching (because it ran too early).
     - Added cartEvent.ts that notify when updating. (event)
- Backend:
  - No changes.

## Does this use the v2 API?:
No. Explicitly using v1 (`{ useV2: false }`). v2 not required because there are no breaking API changes.

No new service-to-service communication. (No C4 L2 update needed.)

**Before:** Cart badge count did not update until a full page refresh.  
**After:** Cart badge updates immediately after add/remove without page reload.  

_Screenshots/GIF to attach:_
image
shows the automatic(live) refresh instead of having to reload the page.

### DIFF
--- a/petclinic-frontend/src/features/inventories/AddInventoryForm.tsx
+++ b/petclinic-frontend/src/features/inventories/AddInventoryForm.tsx
@@ -11,6 +11,13 @@ interface AddInventoryProps {
   refreshInventoryTypes: () => void;
 }
 
+type FieldKey =
+  | 'inventoryName'
+  | 'inventoryType'
+  | 'inventoryDescription'
+  | 'inventoryImage'
+  | 'inventoryBackupImage';
+
 const AddInventoryForm: React.FC<AddInventoryProps> = ({
   showAddInventoryForm,
   handleInventoryClose,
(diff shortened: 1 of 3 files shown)


elPerax (approved):
Hey Olivier, good work here, the auto-refresh on the cart notification dot makes the UX a bit more smooth than before, since the badge updates right away after adding or removing. The code is good and works as expected. Just small things to watch that would be to avoid double-triggering updates, add cleanup if using an effect or listener, and make sure the dot hides correctly at zero. Nothing that blocks my approval though, so I’m going to approve it.

Chronyyx (approved):
Great! Works in local, no complaints.

PR #1094, opened 2025-09-23
bug(CART-CPC-1501): prevent infinite cart fetch loop

shawn-nabizada:
**JIRA:** [link](https://champlainsaintlambert.atlassian.net/browse/CPC-1501?atlOrigin=eyJpIjoiNDY0ODBmNmI5YWNjNGE5MTk0ZTM0ZDZjZmFkNTE2NDYiLCJwIjoiaiJ9)
## Context:
This ticket solves an infinite loop request to the api-gateway. This ticket is important because the loop was flooding the backend with repeated requests, generating infinite logs in the console and consuming unnecessary resources, which could impact performance and make debugging much harder.
No!
## Changes
Only file UserCart.tsx was affected, and in the file I:
- Refactored useEffect dependencies to avoid repeated backend requests
- Moved cart item count update to a separate effect
- Fixed TypeScript errors with updateCartItemCount
- Ensured wishlistUpdated is only reset after fetch
## Does this use the v2 API?:
PR uses v2 API, the file I modified had already been using v2 and I didn't touch it.
No
No new UI changes.


elPerax (approved):
Looks good to me. The change cleanly fixes the infinite fetch loop by limiting the effect dependencies to cartId and wishlistUpdated, and separating the cart count update into its own effect on cartItems. That way the count updates without re-triggering the fetch. The logic is straightforward and the code is easy to follow. So I approve it.

2333476 (approved):
This change looks good and should effectively prevent the infinite cart fetch loop.

PR #1097, opened 2025-09-23
but(VIST-1397): fix api calls in vist

anvsult:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-1397
## Context:
Most api calls on the visits frontend are hardcoded and don't work on the deployed version.
These should be using the shared axios instance.

## Changes

The following changes are supposed to fix the api endpoints not working on the deployed version because they now all use the shared axios instance. The changes only affect the visits service.

- Most changes are there to use the axios instance and specify the api version.

- getEmergencyVisit.ts
This file was renamed to getEmergencyVisitsByOwnerId.ts to make its purpose more clear

- deleteReview.ts
This function was moved to its own file because it was manually implemented in multiple other files

- getReview.ts
This function was moved to its own file because it was defined in the editReview.ts file 

- VisitsListTable.tsx
Some logic refinement was done on this file to make it work with the newly implemented axios instance

## Does this use the v2 API?: Yes
This PR uses V2 for some api endpoints because those endpoints did not work with the V1 version

This PR does not affect the UI

## Dev notes (Optional)
Nothing to be noted

No linked pull requests


MattysLeduc (approved):
The shared axios instance is consistently used across all endpoints. API versioning is explicitly specified, with v1 preferred and v2 only applied where breaking changes are required. Error handling with try-catch blocks is implemented where needed. All URLs now rely on environment variables, with no hardcoded values or manual concatenation.

isaacwallace123:
Maybe fix the title of the PR.... "but" --> "bug"?

jeremy-misola:
but

PR #1099, opened 2025-09-24
feat(INVT-CPC-1493): add more quantity to inventory supply

CunninghamLi:
**JIRA:**
https://champlainsaintlambert.atlassian.net/browse/CPC-1493?atlOrigin=eyJpIjoiMWRlNTRiNjNiNjU4NDJlZmJkOGQ4NmJlMTU4YTNjOWIiLCJwIjoiaiJ9
## Context:
This change addresses ticket INVT-CPC-1493. The goal of the ticket is to allow inventory managers to increase the quantity of an existing supply item directly from the inventory list so that stock levels can be kept accurate when restocking. It also enforces rules to keep the inventory in check. I added an “Add Quantity” button that updates the supply quantity instantly when clicked. When inside the Edit form, the implementation rejects invalid inputs such as 0, negative values, non-numeric values, or decimals. It also enforces the business rule that the maximum quantity is 100  so no more supplies can be added once that limit is reached as well as the minimum being 1 since 0 is rejected. On success, the inventory list updates immediately with the new total so that the user sees the result right away.
No.
## Changes
Added an “Add Quantity” button in the inventory products table.
Implemented logic to increment the supply quantity by +1 while enforcing validation rules (no negative, zero, decimal, or non-numeric quantities, and enforcing the max limit of 100).
Updated the InventoryProducts.tsx component and connected it with the backend endpoint (restockProduct) through axiosInstance.

Other modules that were indirectly affected include:
ConfirmationModal (still used for delete actions, unchanged but coexists with the new button).
Shared API utilities (axiosInstance) since the new endpoint calls go through it.
CSS styles for the inventory table to ensure that the new button fits within the Actions column without breaking the layout.
Applied small error-handling improvements so that users receive feedback if the quantity update fails.

Only 2 files changed:
InvetoryProducts.tsx and InventoryProducts.css
## Does this use the v2 API?:
No.
No.
Before
image
After
image
In the edit form, here are the rules:
Quantity must be greater than 0
image
Quantity can't go above 100
image
You must enter a quantity
image
Add quantity button is non pressable and greyed out if quantity is at 100
image


christianJames24 (dismissed):
Good frontend ticket, it fixed alot of issues all in one and really helps out

OlivierGingras026 (approved):
The changes were made , and even fix an error with the 0 value in the form.

josevmorilla (approved):
Status color was re-added and quantity doesn't go over 0 or 100 with the buttons and forms. 
Good changes 👍

PR #1106, opened 2025-09-24
feat(INVT-1549): add product profit margin to supply list

josevmorilla:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-1549
## Context:
Inventory managers need to be able to see how much profit each product generates in the supply list. This PR calculates the productProfit value so managers can compare cost vs sale price of each supply.
No

## Changes

- Added productProfit field to Product entity and productResponseDTO
- Updated InventoryProducts.tsx to calculate and attach a productProfit field for each product (salePrice - costPrice)
- Updated InventoryProducts.tsx to display cost price, sale price and profit 
- Updated InventoryProducts.css to properly display the new headers and better fit the contents of profit 
- Added a colored based value to display if profits are positive or negative
- Updated the inventory pdf report to display cost, sale price and profit

## Does this use the v2 API?:
No
No
## Before and After UI 
Before: supply list only showed Sale Price.
Screenshot 2025-09-24 183832
After: supply list shows Cost | Sale Price | Profit.
Screenshot 2025-09-24 184353

## Dev notes
Negative profit values are displayed in red to surface loss-making items, albeit this can be a new exception and rule but there are scenarios where a manager would need to sell supply under cost (e.g. bad product, clearing out inventory for new stock)

### DIFF
--- a/.github/workflows/parallel_microservice_build.yml
+++ b/.github/workflows/parallel_microservice_build.yml
@@ -29,8 +29,18 @@ jobs:
       run: chmod +x ./gradlew
       working-directory: ./api-gateway
     - name: Build with Gradle Wrapper
-      run: ./gradlew clean build
+      run: ./gradlew clean build jacocoTestReport
       working-directory: ./api-gateway
+    - name: Verify coverage threshold
+      run: ./gradlew jacocoTestCoverageVerification
+      working-directory: ./api-gateway
+      continue-on-error: true
+    - name: Upload coverage reports
+      uses: actions/upload-artifact@v4
+      with:
+        name: coverage-reports-api-gateway
+        path: ./api-gateway/build/reports/jacoco/test/html/
+        retention-days: 30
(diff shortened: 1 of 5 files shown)


CunninghamLi (approved):
Well done, the addition of this feature enhances the experience because as an inventory manager we can finally see if we are making profits or not. Also the addition of the color green or red based on if the profits are negative or positive is a nice feature. Fixing the PDF to accommodate to these changes was done accordingly and well. Overall,  the UI did not break and I approve of this PR.

OlivierGingras026 (approved):
The addition of a profit margin is a huge addition, making the inventory service more usuable. You did calculation as well with  product.productSalePrice- product.productPrice).toFixed(2) which is good since you did not use hardcoded values to do it. You have also fix the filter issue we had which now make us enable to see even low stock product in the "All" filter which is great.

PR #1107, opened 2025-09-25
feat(CUST-CPC-1537): Setup File Service

megadoxs:
**JIRA:** [link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-1537?atlOrigin=eyJpIjoiMGQwY2U3ZWMxOTlmNDI1MWI2ZGFiYzI3ZDAzODZhODciLCJwIjoiaiJ9)
## Context:
This ticket and PR are about adding a new service that would handle files in general, but more importantly images. A lot of services are either storing their image in the react front end or adding a postgress database to store them. This service aims to centralize the location of image.
No
## Changes
This PR adds a new service entirely written in go that currently only handle the get request.
## Does this use the v2 API?:
No
yes
Doesn't change anything on the front-end


megadoxs:
The port to the service is currently exposed to the outside because I need testing to make that the buckets and image added to it stay stored in the minio. This exposed port will be removed before being merged.

if anybody could please tell me if they get any errors when running the following command, it would be really appreciated.
http://localhost:8000/images/3e5a214b-009d-4a25-9313-344676e6157d

nic5694:
> Adding a volume should make it persistent, but what I am most worried about is if how I preloaded data inside of it will work. Since we do need preloaded image for some service like vets and users.

Yes that is a big concern because if it is not preloaded correctly the endpoints can change and it can mess up services using specific endpoint links. Typically these are used to mock when running locally commonly called multi-cloud gateways to mock S3 and azure blob storage and more. Its also that these are typically meant to run in Kubernetes and docker while we are using it to run production. Is not a typical PROD environment

OlivieriD (approved):
Reviewed and confirmed this story meets the user story of ticket: https://champlainsaintlambert.atlassian.net/browse/CPC-1537

PR has been reviewed and aligns the changes with precedent choices, such as creating a new file service that would handle all future image and file upload/download and store them in the database. The file service also uses dynamic URLs rather than static hard-coded URL, which is good practice.

All acceptance criteria are met:

- The new service must be added to the petclinic directory
- The new service must build successfully.
- The new service must compose on dev and prod
- The new service must handle get file by id.

I approve moving this story to be labeled as "Done".

OlivieriD (approved):

PR #1109, opened 2025-09-25
feat(INVT-CPC-1539): mark inventory as important

christianJames24:
**JIRA:** [link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-1539)
## Context:
This ticket adds an important boolean to the backend of inventory-service, which is then used on the frontend to mark certain inventories as important. A new filtering checkbox has also been added for this new feature which is used to filter, in addition to the previously existing flters, the inventories marked as important. This is for the customer frontend.
No

## Changes
added a new Boolean to the backend called "important"
modified files related to inventories in the api-gateway and inventory-service to accomodate for this change.
added new tests for the new Boolean value and modified some previously existing tests to include it as a null parameter so that they won't fail.

on the frontend, new file: toggleInventoryImportant
modified InventoriesListTable, searchInventories, useSearchInventories, Inventory

in InventoriesListTable:
a star svg was added to each inventory card and a checkbox with important is added to the filtering system
a new function called handleToggleImportant was added and previous code with filters had importantOnly added to it.

in searchInventories:
importantOnly fields were added to it.

## Does this use the v2 API?:
No
No
BEFORE:
no important filter or checkbox:
image

AFTER:
default page now (with a few selected as important):
image
now with important filter checked:
image
in combination of category equipments and important checkbox:
image
in combination of category name "diag" and category equipments and important checkbox
image

## Dev notes (Optional)
in order for this to work I had to fix a filtering bug in useSearchInventories that should've been done in my previous pull request[1091]. everything is fine now however...


OlivierGingras026 (approved):
This PR looks really good and brings clear improvements on both the backend and frontend. On the backend, the new "important" boolean is well integrated into the inventory-service and the api-gateway. The test coverage is solid (JaCoCo above 90%), and it’s nice to see the existing tests updated to handle null values so they won’t break.
On the frontend, the toggleInventoryImportant function built with Axios is simple and effective. Adding the star SVG to the inventory cards is a great usability touch, and the new filtering option makes the feature genuinely useful. Splitting out the toggleInventoryImportant logic also keeps the code clean and maintainable.

Overall, this is a well-implemented feature that improves the user experience without adding extra complexity.

PR #1111, opened 2025-09-26
bug(INVT-CPC-1465): Fix translation formatting

CunninghamLi:
**JIRA:**
https://champlainsaintlambert.atlassian.net/browse/CPC-1465

## Context:
The ticket is about fixing the UI when translating to other languages, specifically when translating into certain languages, some texts don't fully show up which cuts off some words. Also when using the edit or move button, when brought to their respective page, the translation formatting is also bad there. We are doing this change to make the UI cleaner whether it's in english or any other translated language for a better overall user experience.

No.

## Changes
- The only file changed was InventoryProducts.css. 
- Allowed table headers (<th>) and cells (<td>) to wrap/break text so translated French/German/Spanish labels don’t overflow.
- Added column-specific widths (Supply Id, Supply Name, Description, etc.).
- Overrode fixed widths for form buttons like Update and Move so translated labels fit without splitting or clipping.
- Added CSS rules: min-width, white-space: nowrap, and width: auto

## Does this use the v2 API?:
No.

No.

Before(Text is cut off in other languages and is replaced with "..."
Screenshot 2025-09-25 203642
After(Text fully shows)
image
Before(Text is overflowing when translated for edit button)
Screenshot 2025-09-25 205030
After(Fixed)
image
Before(Text is overflowing when translated for move button)
image
After(Fixed)
image

### DIFF
--- a/petclinic-frontend/src/features/inventories/InventoryProducts.css
+++ b/petclinic-frontend/src/features/inventories/InventoryProducts.css
@@ -200,3 +200,43 @@ select:focus {
   margin: 20px 0;
   float: right;
 }
+.table th,
+.table td {
+  white-space: normal;
+  overflow-wrap: anywhere;
+  word-break: normal;
+  text-overflow: initial;
+  line-height: 1.2;
+}
+
+.table td:nth-child(4),
+.table td:nth-child(5),
+.table td:nth-child(6),
+.table td:nth-child(7) {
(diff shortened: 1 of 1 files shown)


OlivierGingras026 (approved):
As Chrisitan said, short but effective. The css brings more quality to the UI.

PR #1121, opened 2025-09-26
Feat/auth cpc 1444 handle expired sessions

AlexRemiNoel:
**JIRA:**https://champlainsaintlambert.atlassian.net/browse/CPC-1444
Context:

When the tocken expires it deletes it when you press the home button

no

Reviewers need to check for any changes to the
.vscode folder and add a comment about it to their review comments.
Changes

only the axios error handler
Does this use the v2 API?:

no

no
Dev notes (Optional)

Quick fix for now might need to implement refresh tokens later

### DIFF
--- a/petclinic-frontend/src/shared/api/axiosErrorResponseHandler.ts
+++ b/petclinic-frontend/src/shared/api/axiosErrorResponseHandler.ts
@@ -17,10 +17,10 @@ export default function axiosErrorResponseHandler(
 ): void {
   const redirectPath = errorPageRedirects[statusCode];
   if (statusCode ==401){
-      console.log("redirect to logout");
+      console.log("clear the tocken");
       localStorage.removeItem("token");
       localStorage.clear()
-      router.navigate('/logout')
+      router.navigate('/home')
   if (redirectPath) {
     // log for easy debug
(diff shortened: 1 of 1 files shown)


jeremy-misola:
if it works it works, at least theres a temporary fix to it so we don't have to open incognito every 2 seconds. I'll make another ticket for integrating an actual token refresh strategy

petclinic-frontend/src/shared/api/axiosErrorResponseHandler.ts
```diff
@@ -16,13 +16,17 @@ export default function axiosErrorResponseHandler(
...
-
-  if (redirectPath) {
-    // log for easy debug
-    console.error(`Redirecting to ${redirectPath} due to error:`, error);
-    router.navigate(redirectPath);
-  } else {
-    // log whatever that wasn't handled
-    console.error('Unhandled error:', error, 'Status code:', statusCode);
+  if (statusCode == 401) {
+    localStorage.removeItem('token');
```
DylanBrass, line 20: There is no token in the local storage

DylanBrass (changes requested):

DylanBrass:
Do not resolve comments that have not been resolved.

jeremy-misola (approved):
looks fine

PR #1128, opened 2025-09-26
feat(INVT-CPC-1550) Change-Inventory-UUID-Into-Inventory-Name 

Seblxx:
**JIRA:** link to jira ticket
https://champlainsaintlambert.atlassian.net/browse/CPC-1550
## Context:
The ticket aims to improve clarity towards the selected Inventory. Previously, we saw the inventory’s ID when we selected inventory instead of it’s name, which is absolutely unecessary and pretty confusing. Now for any inventory that is selected, you can see it’s respective name and know exactly what Inventory you have selected when the data gets really numerous to handle.
No.
## Changes
The files changed were InventoryProducts.tsx I added a component inventoryName and used a useEffect method to map the name of the inventory by matching it to it’s respective InventoryId. I also added an error catcher than says it couldn’t fetch the details right if an error occured. I then simply replaced the InventoryID by InventoryName between the <span> tag.
## Does this use the v2 API?
No.
No.
The previous UI used to display an InventoryID when the inventory was selected, it looked like this:
oldui
After my changes, it now looks like this. Depending on the selected inventory, you'll have the right name assigned to it.
newui


OlivierGingras026 (changes requested):
You have to put your branch inside feat/ now its not

OlivierGingras026 (approved):
Effective change, more readable now

PR #1138, opened 2025-09-27
feat(AUTH-CPC-1522): Remove new emailing service implementations and improve the design of the old mailing service

JeremyNRoos:
**JIRA:** [link](https://champlainsaintlambert.atlassian.net/browse/CPC-1522?atlOrigin=eyJpIjoiMzBkOTllZDFhNzdiNDhmNThmOTg0YmU3MTFhM2E2ZGQiLCJwIjoiaiJ9)
## Context:
Change all the uses of the new emailing service with the go mailing service, as well as improve the design of the old mailing service. This way we keep things simple easy to understand. 
No

## Changes
-Changed the old mailing service's sendEmail method in the UserService and UserServiceImpl
-Added new dto for old mailing service to base everything off of
-Removed new emailing service's methods in UserService and UserServiceImpl
-Refactored the tests in the auth service for the old mailer service that were affected by dto change
-Removed the new emailing service's tests
-Deleted EmailingServiceClient(new emailing service)
-Changed the old mailing service struct to match the new dto in the auth service
-Refactored the files in old mailing service folder in auth service to work with new dto
-Refactored the code in the controller, middleware, service, and mailer in the old mailing service to work with new struct
-Refactored all the tests for the go service to match with the new struct
no
Did not touch front end
## Dev notes (Optional)
I needed to change go version in from 1.20 to 1.25(1.24 should work) for some packages that are imported in the service. I will change it back in the PR. So if the go packages are being found when you try to build the project, change the version from 1.20 to 1.24 or newer.


isaacwallace123:
Maybe try to change NewMail to another name? Like, why is the DTO it called NewMail... it should maybe be something more conventional

auth-service/src/main/java/com/petclinic/authservice/businesslayer/UserServiceImpl.java
```diff
@@ -424,64 +346,15 @@ public void processResetPassword(UserResetPwdWithTokenRequestModel resetRequest)
...
-
-        Mail mail = Mail.builder()
-                .message(email)
-                .subject("PetClinic forgot password")
-                .to(recipientEmail)
-                .build();
-
-        mailService.sendMail(mail);
+
+        NewMail newMail = new NewMail(
```
isaacwallace123, line 350: This is an example of why it should not be called "NewMail" considering new is a reserved word for Java. Something more appropriate would be MailEntity or MailObject so you can have a more ubiquitous language.

I.E:

MailEntity newMail = new MailEntity()

auth-service/src/main/java/com/petclinic/authservice/businesslayer/UserService.java
```diff
@@ -43,9 +43,9 @@ public interface UserService {
 
     List<UserDetails> findAllWithoutPage();
 
-    Mail generateVerificationMail(User user);
+//    NewMail generateVerificationMail(User user);
 
-    void generateVerificationMailWithNewEmailingService(User user);
+    NewMail generateVerificationMail(User user);
```
AmineBaha-oss, line 48: Method name handleMailPOST does not follow Go conventions

Post should not be all in uppercases

AmineBaha-oss (dismissed):
The old emailing service configuration is still present and should be cleaned up.
in :
auth-service/src/main/resources/application.yml
   emailing-service:  
     host: localhost
     port: 5115
Since we will be working with the new emailing going forward
    
To add clarification for tour dev notes :

you should use go-version: "1.24" since its the one we currently have for the github workflows   

Other than the service configuration that is still present and some inconsistent naming everthing looks good !

auth-service/src/main/java/com/petclinic/authservice/businesslayer/UserServiceImpl.java
```diff
@@ -424,64 +346,15 @@ public void processResetPassword(UserResetPwdWithTokenRequestModel resetRequest)
...
-
-        Mail mail = Mail.builder()
-                .message(email)
-                .subject("PetClinic forgot password")
-                .to(recipientEmail)
-                .build();
-
-        mailService.sendMail(mail);
+
+        Mail newMail = new Mail(
```
isaacwallace123, line 350: This is a much better naming convention. Thank you for taking my advice into consideration and applying perfectly. Great job, I approve.

isaacwallace123:
Everything looks good. All the checks worked first try. Good work Jeremy

jeremy-misola (approved):
looks good

PR #1139, opened 2025-09-27
bug/BILL-CPC-649-Fix_Post_Method

MirkoDi5:
Link to Jira ticket: https://champlainsaintlambert.atlassian.net/browse/CPC-649

### Context:
-----------------------------------------------------------------------------------------------------------------------------------------------
What/Why: This ticket is about fixing the post methods so bills can be created and be viewed and adding a negative test method to test post. This was fixed because the post method did not work so bills were not able to be made. The test method was made to ensure if ever a bill status is null a bad request is returned.

### Does this change affect the .vscode folder in the front-end:
-----------------------------------------------------------------------------------------------------------------------------------------------
No

### Changes:
-----------------------------------------------------------------------------------------------------------------------------------------------
Changed the Uri in the pet-clinic front-end application to match the Uri from the bill controller class. From: "/bill/admin" -> To: /bill. Added a negative test for post to test a null bill status.

### Does This use the v2 API:
-----------------------------------------------------------------------------------------------------------------------------------------------
No

### Does this add a new communication between services:
-----------------------------------------------------------------------------------------------------------------------------------------------
No

### Before/After Ui changes:
N/A


TMBJake (approved):
Thanks for fixing it.

PR #1156, opened 2025-09-27
 feat(AUTH-CPC-839): Change Username Backend No FE

AlexRemiNoel:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-839
## Context:
This code is the same as the previous one it works, however it has no frontend to not conflict with other teams.
no

## Changes
implements the backend to change username
## Does this use the v2 API?:
yes since their used to be the same feature but its commented out
no


isaacwallace123 (approved):
Seems relatively simple. Upon testing, this does work however a front-end for this would be nice to have.

JeremyNRoos (approved):
This is a nice addition to have, nice work!

jeremy-misola:
> Seems relatively simple. Upon testing, this does work however a front-end for this would be nice to have.

real

jeremy-misola (commented):
looks fine

PR #1162, opened 2025-09-28
bug(AUTH-CPC-1598): Homepage contact band buttons are misplaced

isaacwallace123:
**JIRA:** [link](https://champlainsaintlambert.atlassian.net/browse/CPC-1598)

## Context:

This ticket simple fixes the slight styling issue with the home page where the contact band had misplaced components. This was simply a bug that I thought needed to be addressed in this specific sprint. I also took this opportunity to spice up the FAQ page with a small additional animation.

No changes to `.vscode`.

## Changes

- **Home:** Fixed the contact band by fixing column seperation and adding better bootstrap classes + xs.
- **FAQ:** Added more animation using the Reveal component created in a separate ticket.

## Does this use the v2 API?:

No changes to any API calls.

No. Reuses existing endpoints; no new service-to-service calls. (No C4 L2 updates required.)

**Before:**

Before

**After:**

After

### DIFF
--- a/petclinic-frontend/src/pages/FAQ/FAQ.tsx
+++ b/petclinic-frontend/src/pages/FAQ/FAQ.tsx
@@ -1,10 +1,10 @@
-// src/pages/FAQ.tsx
 import { Container, Row, Col, Accordion, Form, Button } from 'react-bootstrap';
 import { NavBar } from '@/layouts/AppNavBar';
 import { AppFooter } from '@/layouts/AppFooter';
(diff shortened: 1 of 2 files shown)


Jeefos (approved):
Very nice change, it looks way better then before where it was all chunky.

JeremyNRoos (approved):
That makes it so much cleaner. I really like it!

isaacwallace123:
> Very nice change, it looks way better then before where it was all chunky.

I know right??? I don't know what I was thinking of when I made it so damn big

PR #1169, opened 2025-09-28
Feat(INVT-CPC-1585): Add cancel button to edit form

J-DPAL:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-1585

## Context:
This change addresses ticket INVT-CPC-1585. The goal of the ticket is to add a Cancel button to the inventory edit form so that users can discard their changes and restore the original inventory values. This ensures users don’t mistakenly update inventory information and provides a safer editing workflow.

No.

## Changes

Added a Cancel button to the Inventory Edit Form.

Stored original inventory values (name, type, description, image, etc.) on initial load so that the cancel action can restore them.

Implemented handleCancel logic to reset form fields to the original values instead of keeping unsaved edits.

Updated InventoryEditForm.tsx (and related state handling) to support the new cancel functionality.

## Does this use the v2 API?:
No.

No.

## Before and After UI 
Before:
The edit form only allowed users to save their changes or exit the form manually.

After:
New Cancel button is visible on the edit form.

Clicking Cancel restores all fields (name, type, description, image) back to their original values and exits the form.

Screenshot Before:
image

Screenshot After:
image

### DIFF
--- a/billing-service/src/test/java/com/petclinic/billing/businesslayer/BillServiceImplTest.java
+++ b/billing-service/src/test/java/com/petclinic/billing/businesslayer/BillServiceImplTest.java
@@ -867,7 +867,132 @@ public class BillServiceImplTest {
                 .verify();
     }
 
+    @Test
+    void getNumberOfBillsWithFilters_Positive_ShouldCountMatchingBills() {
 
+        Bill b1 = buildBill();
+        b1.setBillId("B-1");
+        b1.setCustomerId("C-1");
+        b1.setOwnerFirstName("Alice");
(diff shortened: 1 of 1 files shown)


OlivierGingras026 (approved):
Great improvement! Adding the Cancel button significantly enhances the UX by preventing accidental updates. Storing the original inventory values on load and restoring them through the handleCancel logic is a clean solution that keeps state management predictable. The changes to InventoryEditForm.tsx are straightforward and maintainable, and the workflow now feels much safer for users editing inventory data.

PR #1172, opened 2025-09-28
Bug(INVT-CPC-1601): Fix Get All Types In Angular Frontend

OlivierGingras026:
JIRA: link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-1601)

Context:

The ticket is about allowing when you add an inventory to be able to choose an inventory type. It was needed because in the Angular FE it was not possible to fetch the endpoint to get all types in the add inventory form. It was also necessary to fix the update inventory endpoint in the react FE because the url was not following the conventions.

No.

Reviewers need to check for any changes to the .vscode folder and add a comment about it to their review comments.

Changes

1. Change the url in inventory-form-controller.js to $http.get("api/gateway/inventories/types")
2. Change the url of EditInventory = 'inventories/inventory/:inventoryId/edit' to EditInventory = 'inventories/:inventoryId/edit' in path.route.ts

Does this use the v2 API?:
No.

No.

Before in the Angular FE: 
Screenshot 2025-09-28 125858

After in the Angular FE:
Screenshot 2025-09-28 131541

Before in the React for the update form(2 times the word inventory in the url): 

Screenshot 2025-09-28 140309

After in the React (cleaner url):

image

Dev notes (Optional)

### DIFF
--- a/api-gateway/src/main/java/com/petclinic/bffapigateway/presentationlayer/v1/InventoryControllerV1.java
+++ b/api-gateway/src/main/java/com/petclinic/bffapigateway/presentationlayer/v1/InventoryControllerV1.java
@@ -99,7 +99,7 @@ public class InventoryControllerV1 {
 
 
     @SecuredEndpoint(allowedRoles = {Roles.ADMIN, Roles.INVENTORY_MANAGER})
-    @GetMapping(value = "{inventoryId}", produces = MediaType.APPLICATION_JSON_VALUE)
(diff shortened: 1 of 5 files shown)


christianJames24:
good change, the changes were minimal but got the job done, which is what i like to see

josevmorilla (approved):
solid changes, this is useful for when we're gonna work on the angular fe

PR #1173, opened 2025-09-28
bug(VETS-CPC-1018): Use DTO for vet photos instead of Resource

maxkoryakin:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-1018
## Context:
Updated the photo endpoints to use Data Transfer Objects (DTOs) for internal service communication while maintaining backward compatibility at the API Gateway level. The implementation is used between microservices, but the external API maintains its original contract.

## Changes
Backend (vet-service):

- Created PhotoRequestDTO
- GET, POST, and PUT requests now accept and return PhotoResponseDTO
- Removed overloaded methods that accepted 'MultipartFile' and 'byte[]'

Backend (api-gateway):

- Created PhotoRequestDTO
- VetsServiceClient now handles all the DTO conversions internally
- In PhotoServiceImpl updated all methods to work with DTOs. Uses `EntityDtoUtil.toPhotoResponseDTO()` for converting entities to DTOs and handles content type determination from DTO fields

Frontend (petclinic-frontend):

- No changes made

## Does this use the v2 API?:
No
No

- No changes made to the UI

## Dev notes (Optional)


tamimafg6 (approved):
As Logan mentioned, everything works fine. Also, after checking the code, it looks cleaner

PR #1176, opened 2025-09-28
bug/(VISIT-CPC-1332): 400-error-owner

MariaIsabelGuerrero:
JIRA: https://champlainsaintlambert.atlassian.net/browse/CPC-1332
Context:
This ticket fixes a 400 Bad Request error that occurs when owners attempt to book visits through the frontend. The issue was caused by the frontend sending an extra visitStartDate field in the API request payload, which the backend VisitRequestDTO doesn't expect. The backend only accepts the visitDate field with the format yyyy-MM-dd HH:mm.

Changes:
Owner booking visit component: Removed the spread operator ...visit from the formattedVisit object creation to prevent sending the unwanted visitStartDate field
Explicitly specified only the required fields (visitDate, description, petId, practitionerId, status) that match the backend VisitRequestDTO expectations
Maintained the ownerId and jwtToken fields required for owner authentication and booking functionality

Does this use the v2 API?:
Yes, both the add and update visit functions use the v2 API ({ useV2: true }). This is consistent with the existing codebase implementation for visit operations.
No, this PR does not add new service communications. It fixes existing communication between the frontend and the visits service through the API gateway.
This is not a UI change - it's a backend communication fix. The UI remains the same, but the functionality is now working properly:
Dev notes (Optional)

Linked pull requests (Optional)
None


MattysLeduc (approved):
The changes added are working

anvsult (approved):
Looks good to me

PR #1187, opened 2025-09-29
 fix(VIST-CPC-1567): Fix CSS for Visits Front Page Tables

Pkdotts:
**JIRA:** https://champlainsaintlambert.atlassian.net/browse/CPC-1567
## Context:
No.

## Changes
This PR changes the look of the admin panel for the visits page. This mainly changes the CSS for the tables as well as the layout, as well as adding buttons. This PR also fixes the search bar feature to make it functional. This PR also fixes the header of the Emergencies page.
## Does this use the v2 API?:
No.
No.
Before
image
image

After
image


anvsult (changes requested):
Apart from the hardcoded api, it looks good

anvsult (approved):
Thank you for fixing the hard coded api

MattysLeduc (approved):
Thank you for changing the hard coded URL

PR #1196, opened 2025-10-01
bug(VIST-CPC-1519): refractored endpoints to V1

MattysLeduc:
**JIRA:** [link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-1519)
## Context:
This PR focuses on refactoring and stabilizing the visits module across the API Gateway and frontend. The goal is to centralize all visit-related logic into dedicated controllers, ensure frontend calls are consistent with backend endpoints, and improve maintainability by consolidating tests.
No

## Changes

Refactored visits endpoints

- Moved all visit-related endpoints into a new controller in the new V1 folder so they can be used by everyone.

- Added the breaking v2 endpoints into the same controller to ensure backward compatibility and avoid breaking changes.

Frontend fixes

- Change all methods to use V1 (useV2: false)

Test refactoring

- Consolidated all visit-related tests into the VisitsControllerV2UnitTest class.

- Updated the test structure to align with the new controllers and API endpoints.

## Does this use the v2 API?:
No, after refactoring, all the methods are now using V1 (useV2: false)
no
No changes
## Dev notes (Optional)
N/A
N/A

### DIFF
--- a/api-gateway/src/main/java/com/petclinic/bffapigateway/presentationlayer/BFFApiGatewayController.java
+++ b/api-gateway/src/main/java/com/petclinic/bffapigateway/presentationlayer/BFFApiGatewayController.java
@@ -345,149 +345,149 @@ public class BFFApiGatewayController {
 
 */
 
-        /* Visits Methods */
-
-    @SecuredEndpoint(allowedRoles = {Roles.ADMIN})
-    @GetMapping(value = "reviews", produces = MediaType.TEXT_EVENT_STREAM_VALUE)
-    public Flux<ReviewResponseDTO> getAllReviews(){
-        return visitsServiceClient.getAllReviews();
-    }
-
-
-
-
-
-
-
-
-
-
-    @SecuredEndpoint(allowedRoles = {Roles.ADMIN})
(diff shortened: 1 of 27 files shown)


anvsult (approved):

anvsult (changes requested):
As mentioned in the comments, simply fixing the typo and using notFound instead of badRequest when appropriate

MattysLeduc:
> As mentioned in the comments, simply fixing the typo and using notFound instead of badRequest when appropriate

You are right, I will change it where it is appropriate.

MattysLeduc:
> Hey, I got a couple of questions:
> 
> * Should the tests be fixed instead of being commented out?
> * In `VisitListTable.tsx` (around line 161), do we actually need `api/gateway`?
> * Are we able to use the axios instance instead of EventSource?

The tests inside BFF are no longer needed because all controller methods are now in a new folder. I have made tests for this new class, which is VisitsControllerV1.

For the EventSource and the URL, it will be fixed in another ticket, which is CPC-1698

ClaytonTKC (approved):

aPerson62682 (approved):

PR #1223, opened 2025-10-03
bug(VETS-CPC-1629): Fix the API calls on the vet details page

maxkoryakin:
**JIRA:** [link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-1629)
## Context:
There used to be a bug with the API calls where certain requests to the backend were using the v1 API instead of the v2, which was breaking the requests.

No

## Changes
- Changed certain axiosInstance API requests to use the correct and working version of the API (v2 instead of v1)
- Brought back certain permissions that were commented out in the previous code, which fixes every permission-based request to the API (e.g. load the vet profile picture). 

## Does this use the v2 API?:
Yes, it uses the v2 API because there are no existing endpoints currently that work in the same way on the v1 API. 

No

No changes made to the UI


wolvy27 (changes requested):
Tested locally on my machine, for some reason when I first build petclinic on my machine and enter any of the vets pages (as an admin) the photo album displays no photos (and continues as I leave the page and enter any of the other vet's pages, the photo album displays no photos). But the moment I refresh the page the album photos are back on every vet page.
Before refresh:
image
After refresh:
image

wolvy27 (approved):
Tested locally on my machine after going on a walk, it works now, probably some local issue on my side.

yacineihaddadene (approved):

PR #1433, opened 2025-10-27
feat(BILL-CPC-1717): Fixed UI pagination

SeanBedard:
**JIRA:** [link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-1717)

## Context:
-As a customer/pet owner and admin I want to have an enjoyable experience with the site’s ui.
-As an admin when i filter the bills i don’t want there to be the same amount of pages as when i get all bills.

No.

## Changes
-Fixed the pagination where some pages had only 3 Bills showing and some had 10(the max).
-Added pagination to the customer bills table.
-Added an option to copy customer id and bill id.
-Added a "clear" button to the search option on admin side.
-Fixed the alignment for the pagination control.
-Added the option to go back to page 1 after going to page 3 and beyond

## Does this use the v2 API?:
Yes because almost all our endpoints are on V2

No.

Before:
image
image

After:
image
image
image

## Dev notes (Optional)
The data in the screenshots are the same on every page because my docker and front end were not running, the page was still working because it wasn't refreshed but the data was bugged.


TMBJake (changes requested):
I cannot let a 22,000 line of change be merged 1h and a half before the window closes. No pr should be this large

MirkoDi5 (changes requested):
I don't think there should be 22k lines changed

PR #1468, opened 2026-08-18
feat(CPC): add workflow to create images

DylanBrass:
**JIRA:** link to jira ticket
## Context:

## Changes
## Does this use the v2 API?:
## Dev notes (Optional)
Pull request links


cgerard321 (approved):

petclinic-frontend/Dockerfile
```diff
@@ -1,39 +1,35 @@
...
 RUN chown -R nginx:0 /usr/share/nginx/html && \
     chmod -R g+r+rwX /usr/share/nginx/html
-
 COPY --from=build /app/dist /usr/share/nginx/html
 # Copy the default Nginx configuration file
 COPY nginx.conf /etc/nginx/nginx.conf
-
 # Expose port 3000 for the NGINX web server
 EXPOSE 3000
-
```
nic5694, line 37: run from userid 10000 or smtg high to not have your containers run privileged. when you dont specify one it runs as userid 0

DylanBrass: Will use security context

nic5694 (approved):
thx for the change

PR #1481, opened 2026-09-19
bug(CPC-1951): Add loading state to useAuthRedirect and ProtectedRouteProps hook

DylanBrass:
**JIRA:** [link to jira ticket](https://champlainsaintlambert.atlassian.net/browse/CPC-1951)

## Context:
When doing an action in the customer portal, especially one's that refresh the page, a race condition would cause the user to be redirected to the 401 (and 403 if 401 was disabled).

## Changes
- Added a wait for isLoading in useAuthRedirect and ProtectedRouteProps
- Added a check to avoid infinite loops of redirect, now if the redirect is the same as the page, we do nothing.


cgerard321:
I think the type should be bug and not fix. 😉

DylanBrass:
> I think the type should be bug and not fix. 😉

Indeed ! :sweat_smile:

petclinic-frontend/src/shared/api/axiosErrorResponseHandler.ts
```diff
@@ -37,12 +37,14 @@ export default function axiosErrorResponseHandler(
...
-    if (typeof window !== 'undefined') {
+    if (
+      typeof window !== 'undefined' &&
+      window.location.pathname !== redirectPath
+    ) {
+      console.error(
+        `Redirecting to ${redirectPath} due to a server error:`,
+        error
+      );
       window.location.href = redirectPath;
```
Eric909090: NitPick -->The highlighted line causes full page reload (makes the application slower)

Eric909090 (approved):
Everything is functional.

petclinic-frontend/src/shared/api/axiosErrorResponseHandler.ts
DylanBrass: Can you give me some details as to when this happens ?

DylanBrass: After some digging I think I understand what you mean. The full reload was already there before this PR (the window.location.href assignment is unchanged). This change only adds a condition that we don't redirect when we are already on the page of the redirect target, so it reduces reloads. Switching to router-based navigation would be the next step, but this is out of the scope of this PR, and would need to be done across the App. I opened a ticket [here](https://champlainsaintlambert.atlassian.net/browse/CPC-2000)

Eric909090: Perfect let me know if you need any other information

DylanBrass: If a comment as been resolved, the commenter normally resolves it. So I will let you resolve it :)

Eric909090: Okay I will speak about it to my team for next Sprint

DylanBrass: I meant the Github comment haha

DanMoraru (approved):
The authentication service is functional and the intended bug fix works.

KylianLaw (approved):
I reviewed and tested the files and everything seem to work on my side.

marwamesk (commented):
Thank you for the fix. I reviewed it

marwamesk (approved):

